Issue #3, opened 2018-08-06
why not support for cpu cluster?


banbanandroid:
I dont't have a gpu cluster but cpu, and why not support for cpu cluster?

cheyang:
It can also support cpu cluster. You don't have to specify `--gpu` when you don't need GPU.

xiechengsheng:
But in cpu cluster we couldn't see any cpu information(total/allocated) in `top` command :scream:

cheyang:
Thanks for reporting it. It will support later. And your contributions are also welcome!

xiechengsheng:
Hi @cheyang , I have deployed arena successfully on my small cpu cluster, but the `tf-job` always failed on my cluster, here is the reproduce step:
```
1. check the cluster status
# ./arena top node
INFO[0000] Use specified kubeconfig file /etc/kubernetes/admin.conf 
NAME      IPADDRESS     ROLE    GPU(Total)  GPU(Allocated)
k8s-mst   192.168.1.60  master  0           0
k8s-nod1  192.168.1.61  worker  0           0
k8s-nod2  192.168.1.62  worker  0           0

2. run one tf job
# ./arena submit tf --name=tf-git5 --image=tensorflow/tensorflow:1.5.0-devel --syncMode=git --syncSource=https://github.com/cheyang/tensorflow-sample-code.git "python code/tensorflow-sample-code/tfjob/docker/mnist/main.py --max_steps 10000"
INFO[0000] Use specified kubeconfig file /etc/kubernetes/admin.conf 

NAME:   tf-git5
LAST DEPLOYED: Tue Aug  7 16:31:56 2018
NAMESPACE: default
STATUS: DEPLOYED

RESOURCES:
==> v1alpha2/TFJob
NAME           AGE
tf-git5-tfjob  0s

3. check the job status
./arena get tf-git5
INFO[0000] Use specified kubeconfig file /etc/kubernetes/admin.conf 
NAME     STATUS  TRAINER  AGE  INSTANCE                NODE
tf-git5  FAILED  TFJOB    0s   tf-git5-tfjob-worker-0  N/A

4. can't see the detailed logs because the pod status is failed.
# ./arena logs tf-git5
INFO[0000] Use specified kubeconfig file /etc/kubernetes/admin.conf 
pod 'tf-git5-tfjob-worker-0' has not been started.

```
5. it seems that arena hasn't implemented describe command yet, so I can't get concrete failing reason. :cry:
[...]

PR #7, opened 2018-08-07
Feature: add cpu/mem info in top node side

xiechengsheng:
Signed-off-by: xiechengsheng <XIE1995@whut.edu.cn>
Fix part of #3 , codes about `top job` would come soon.


cmd/arena/commands/resource.go
```diff
@@ -77,3 +77,71 @@ func gpuInContainerDeprecated(container v1.Container) int64 {
...
+		if gpuInPod(pod) > 0 || cpuInPod(pod) > 0 || memoryInPod(pod) > 0 {
+			podsWithResource = append(podsWithResource, pod)
+		}
+	}
+	return podsWithResource
+}
+
+// The way to get CPU Count of Node
+func cpuInNode(node v1.Node) int64 {
+	val, ok := node.Status.Capacity[v1.ResourceCPU]
```
GarnettWang: this should be `node.Status.Allocatable[v1.ResourceCPU]`

cmd/arena/commands/resource.go
```diff
@@ -77,3 +77,71 @@ func gpuInContainerDeprecated(container v1.Container) int64 {
...
+	containers := pod.Spec.Containers
+	for _, container := range containers {
+		cpuCount += cpuInContainer(container)
+	}
+
+	return cpuCount
+}
+
+func cpuInContainer(container v1.Container) int64 {
+	val, ok := container.Resources.Limits[v1.ResourceCPU]
```
GarnettWang: use container.Resources.Requests[v1.ResourceCPU], not Limits. 

Please also check memoryInContainer.

cmd/arena/commands/resource.go
xiechengsheng: OK, Got it, so the gpu resource function `gpuInNode` and `gpuInContainer` should also be updated?

GarnettWang: no need to update gpu, as gpu resquests equals to limits always.

xiechengsheng: ok, thanks!

xiechengsheng: Emmm, in my opinion the `Allocatable` means node remaining resource, but here we want to get node's total resource, so we should use `Capacity` here instead of `Allocatable`, right?

cmd/arena/commands/resource.go
xiechengsheng: Yes, we should fix this, thanks.

cmd/arena/commands/resource.go
GarnettWang: users should care more about the total resources they can use, not include the reserved resources. it's my opinion.  :)

what's your thought? @cheyang

xiechengsheng:
@GarnettWang codes have been updated, PTAL, thanks~

cmd/arena/commands/resource.go
cheyang: My suggestion is that we should keep consistent with `kubectl describe no`. Take a look at 
https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/describe.go#L2773

xiechengsheng: Hi @cheyang In `kubectl describe no`, it seems that the `Capacity` and `Allocatable` are all showed in response message.

cheyang: We should use the same algorithm of `describe node` when getting the results of `total` and `allocated`.  I think we can leverage their implementation, and make sure our result is consistent with it.

cheyang:
@xiechengsheng , thanks for your quick and great contribution!
I suggest we should get the same result with `kubectl top node`. 

```
kubectl top node
NAME                     CPU(cores)   CPU%      MEMORY(bytes)   MEMORY%
i-bp10ql620010exb478op   166m         4%        2237Mi          28%
i-bp14vuwrktdfxioibi8i   192m         4%        2124Mi          27%
i-bp1afgfln76507diuvl9   234m         5%        2222Mi          28%
i-bp1fzgva3kl983om6iwy   61m          3%        2772Mi          40%
i-bp1fzgva3kl983om6iwz   41m          2%        2358Mi          34%
```
Can you help investigate and implement it? Thanks!

xiechengsheng:
@cheyang OK, It's my pleasure to take this. BTW, do you mean that we should get and output the real GPU/CPU/mem usage in cluster node, instead of `pod request resource value`/`node total resource value`?

cheyang:
For CPU/Memory, I think it should be the same implementation as https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/top_node.go. But for GPU,  we don't need change. Thanks.

cheyang:
>I'm sorry to ask that which metrics server would be used to monitor the whole cluster's resource usage info? heapster or metrics-server? Because different metrics server will make me to write different client code to get cluster info from metrics server's RESTFul API. Thanks a lot.

According to the implementation in [top_node.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/top_node.go), we should use heapster. If heapster is not available in the cluster, we will not display the cpu/memory  usage.

 >And should we really print the real-time resource usage in top node command? This resource usage changes all the time.

According to the definition of `top`, I think it's for real-time resource usage.

xiechengsheng:
Hi @cheyang , sorry for my late updated codes. Please re-check this pr.
BTW, I run `kubectl top pod --all-namespaces` on k8s master to get whole cluster's namespace pods cpu resource usage, and I simply add the each pod's cpu usage to get cluster's cpu sum usage. Then I run `kubectl top node` to get whole cluster's node cpu resource usage, and I add each node's cpu uasge, I find that `sum(node_cpu_usage)` is always bigger than `sum(pod_cpu_usage)`. These two ways to get cpu/mem usage would get different result, then which one do you think is better to be adopted in `arena top node` and `arena top node -d` command?

cmd/arena/commands/heapster.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+var httpClient = &http.Client{Timeout: 30 * time.Second}
+
+func getHeapsterEndpointUrl(client *kubernetes.Clientset) (string, error) {
+	heapsterService, err := client.CoreV1().Services(metav1.NamespaceSystem).Get(DefaultHeapsterServiceName, metav1.GetOptions{})
+	if err != nil {
+		return "", err
+	}
+	heapsterEndpointUrl := DefaultHeapsterScheme + "://" + heapsterService.Spec.ClusterIP + ":" + DefaultHeapsterPort + HeapsterApiPath
```
xiaozhouX: `heapsterService.Spec.ClusterIP` Should be ApiServer IP .
If arena execute outside cluster nodes, it can't reach the heapster ClusterIP.

cmd/arena/commands/top_node.go
```diff
@@ -151,23 +175,56 @@ func displayTopNodeSummary(nodeInfos []NodeInfo) {
...
+		totalCPUsInCluster     int64
+		usedCPUsInCluster      int64
+		totalMemoryInCluster   int64
+		usedMemoryInCluster    int64
 	)
 
 	fmt.Fprintf(w, "\n")
 	for _, nodeInfo := range nodeInfos {
 		totalGPU, allocatedGPU := calculateNodeGPU(nodeInfo)
+		totalCPU, usedCPU := calculateNodeCPU(heapsterEndpointUrl, nodeInfo)
```
xiaozhouX: This will request to heapster for (node num * pod num) times. If we want to get real-time metric, maybe we can reference the Implementation of `kubectl top node / kubectl top pod`
https://github.com/kubernetes/kubernetes/blob/029bb4e213502268af6222cc2f6903ab90de751b/pkg/kubectl/metricsutil/metrics_client.go#L67

cmd/arena/commands/heapster.go
xiechengsheng: Yes, you are right, we should consider this.

cmd/arena/commands/top_node.go
xiechengsheng: Oh, thanks for your kind reminding, then we could just do nothing but include this `metrics_client.go` into our `commands` package~

xiechengsheng:
Hi there, when I add "k8s.io/metrics" in `Gopkg.toml`, and use `dep` to generate dependent package in `vendor` directoory, but the bulid process `make` would always fail with `undefined error` in my server, I have no idea to fix this, need yours help...

Here is the `make` log output:
image

And I think maybe a version error exist here, so I change the "k8s.io/metrics" from branch `release-1.9` to `release-1.10` to `release-1.11`, but it seems useless...

xiechengsheng:
Any help would be appreciated!

xiechengsheng:
I signed it!

xiaozhouX:
@xiechengsheng  Sorry for late response. I exec `arena top node` in my host for a 1.11.x version k8s cluster. It turns out a error `the server could not find the requested resource (get services http:heapster:)`, it seems that there is compatibility problem, Can you check for that?  Thanks very much for your contribution.

PR #22, opened 2018-08-18
generate document for CLI

cheyang:


wsxiaozhang:
can we add some explanation to "namespace", like "the namespace of specific job"  ?

cheyang:
@wsxiaozhang /PTAL

PR #25, opened 2018-08-20
support pod affinity for tfjob

### LINKED ISSUE #24: Make TFJob in bin pack mode
Need to schedule the workers and PS in binpack mode rather spread as default mode. [TFJob #781](https://github.com/kubeflow/tf-operator/issues/781#issuecomment-414180859)

cheyang:
Fix #24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


wsxiaozhang:
/lgtm
let's make it default as false for now. and see whether such kind of ps/worker affinity is valuable to real users.

PR #38, opened 2018-08-29
Fix Path Error

xieydd:
Fix the path error of  `Run a distributed TensorFlow training job with gang scheduler` demo.


cheyang:
Thanks very much!

PR #41, opened 2018-08-30
Show event if pod is pending in get command

xiaozhouX:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/41)


xiaozhouX:
I signed it!

PR #44, opened 2018-08-31
show tfjob events, when has pods pending

xiaozhouX:
when tfjob has pod in pending status, tfjob is still `Running`, show pods events in that status

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/44)

### DIFF
--- a/cmd/arena/commands/get.go
+++ b/cmd/arena/commands/get.go
@@ -167,6 +167,8 @@ func printEvents(w io.Writer, namespace string, pods []v1.Pod) {
 	}
 }
 
+// Get real job status
+// WHen has pods being pending, tfJob still show in Running state, it should be Pending
 func GetJobRealStatus(job TrainingJob) string {
 	hasPendingPod := false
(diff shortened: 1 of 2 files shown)


cmd/arena/commands/get.go
```diff
@@ -167,6 +167,22 @@ func printEvents(w io.Writer, namespace string, pods []v1.Pod) {
...
 }
 
+func GetJobRealStatus(job TrainingJob) string {
+	findPendingPod := false
+	jobStatus := job.GetStatus()
+	if jobStatus == "RUNNING" {
+		pods := job.AllPods()
+		for _, pod := range pods {
+			if pod.Status.Phase == v1.PodPending {
+				findPendingPod = true
```
cheyang: add break here

cheyang: findPendingPod -> hasPendingPod

xiaozhouX: ok, u are right

PR #58, opened 2018-09-19
Support GKE

mrkm4ntr:
Support GKE.
`vendor/k8s.io/client-go/plugin/*` is not included in the source tree, so I ran `dep ensure`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/58)


cheyang:
Thank you for your contribution. Sorry, I'm not familiar with GKE. Could you please clarify the reason of adding GKE vendor? Did it cause arena fail in GKE? Can you explain a little more. Thanks very much.

mrkm4ntr:
If there isn't [this line](https://github.com/kubeflow/arena/blob/d53d3f641d859a385d68a4089457e2bd59cc0666/cmd/arena/commands/root.go#L25), it occurs following error.

```
$ arena list
FATA[0000] No Auth Provider found for name "gcp"
```

More information is [here](https://github.com/kubernetes/client-go/issues/242).
GKE modules are added to the vendor by adding above import.

cheyang:
Understand! Thanks very much.

PR #64, opened 2018-10-10
fix serve submit gpuCount

xiaozhouX:
fix submit serving gpu parameter

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/64)


cheyang:
/lgtm
/approve

Thanks for fixing it!

PR #85, opened 2018-12-07
Check the tfjob status instead of get it

chaoleili:
This better to check the tfjob status instead of get it with
hasCondition. When get tfjobs form kube, sort the tfjobs status
conditions at first and put the newest condition at first. Then
check the status directly form sorted conditions.

Using hasCondition means if the tfjob has the condition at any time,
it alwayse will return True, but not the newest condition of 
the tfjob. So i think check tfjob status will be better and more accurate.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/85)

### DIFF
--- a/cmd/arena/commands/trainer_tensorflow.go
+++ b/cmd/arena/commands/trainer_tensorflow.go
@@ -16,6 +16,7 @@ package commands
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/kubeflow/arena/util"
 	"github.com/kubeflow/tf-operator/pkg/client/clientset/versioned"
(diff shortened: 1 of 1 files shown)


cheyang:
Thank you very much for your contribution! I really like your ideas!  

Would you please do me a favor to change the status to the uppercase. For example,  turn "Running" to "RUNNING".   This can make the status consistent with other type of the job. Thanks again!

chaoleili:
Updated~

cheyang:
Thanks again, @chaoleili . Look forward to your next contribution!

PR #93, opened 2018-12-31
customize cpu profile samples

cheyang:


wsxiaozhang:
the problem is there are multiple times query for trainers during 'list' command, which wastes time obviously. this fix is to use a internal flag 'useCache' to indicate for 'list' command only execute the full query of all types of trainer once.

PR #115, opened 2019-01-27
Support listing training job 

cheyang:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/115)

### DIFF
--- a/cmd/arena/commands/delete.go
+++ b/cmd/arena/commands/delete.go
@@ -55,7 +55,7 @@ func NewDeleteCommand() *cobra.Command {
 			for _, jobName := range args {
 				err = deleteTrainingJob(jobName)
 				if err != nil {
-					fmt.Printf("Failed to delete %s due to %v\n", jobName, err)
+					fmt.Printf("Failed to delete %s, reason is that %v\n", jobName, err)
 				}
 			}
 		},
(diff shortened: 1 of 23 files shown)


cmd/arena/commands/delete.go
```diff
@@ -75,13 +75,19 @@ func deleteTrainingJob(jobName string) error {
 	log.Debugf("it didn't deleted by helm due to %v", err)
```
wsxiaozhang:
```suggestion
	log.Debugf("%s wasn't deleted by helm due to %v", jobName, err)
```

cmd/arena/commands/delete.go
```diff
@@ -75,13 +75,19 @@ func deleteTrainingJob(jobName string) error {
...
-	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+	trainingTypes := getTrainingTypes(jobName, namespace)
+	if len(trainingTypes) == 0 {
+		return fmt.Errorf("There is no training job found with the name %s, please check it with `arena list | grep %s`",
+			jobName,
+			jobName)
+	} else if len(trainingTypes) > 1 {
+		return fmt.Errorf("There are more than 1 training jobs with the same name %s, please check it with `arena list | grep %s`. And use `arena delete %s --type`",
```
wsxiaozhang:
```suggestion
		return fmt.Errorf("There are more than 1 training jobs with the same name %s, please double check with `arena list | grep %s`. And use `arena delete %s --type` to delete the exact one.",
```

PR #118, opened 2019-02-01
fix get gpu metric error & typo error of 'gpu exporter'

soolaugust:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/118)


cmd/arena/commands/gpu_info.go
```diff
@@ -163,8 +163,8 @@ func QueryMetricByPrometheus(client *kubernetes.Clientset, prometheusServiceName
 		return gpuMetric, fmt.Errorf("failed to query prometheus, status: %s", metricResponse.Status)
 	}
 	if len(metricResponse.Data.Result) == 0 {
-		log.Errorf("gpu metric is not exist in prometheus for query  %s", query)
-		return gpuMetric, fmt.Errorf("gpu metric is not exist in prometheusfor query  %s", query)
+		log.Debugf("gpu metric is not exist in prometheus for query  %s", query)
+		return gpuMetric, nil
```
cheyang: May I ask the reason to not raise up the error ?

soolaugust: Actually with local test,  when job is not running(pending), gpu metric is not detected. using `arena top job` command at this time will raise a error. This is not user friendly, so I think it should not be a error message.

cheyang: Okay， Thank you very much for your explanation, testing and contribution! Look forward to your next PR.

PR #178, opened 2019-06-06
Basic struct for arena cli update to support volcano

SrinivasChilveri:
This is the initial PR to support volcano in arena cli.

If we want to test timebeing we can install the back end following the belowi links
https://github.com/volcano-sh/volcano
https://volcano.sh/blog/quick-start-volcano/

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/178)

PR #180, opened 2019-06-14
volcano deployment in kubernetes using helm and charts

SrinivasChilveri:
volcano deployment in kubernetes using helm and charts.

This PR has related to the https://github.com/kubeflow/arena/pull/178  but no dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/180)

### DIFF
--- a/kubernetes-artifacts/volcano-operator/.helmignore
+++ b/kubernetes-artifacts/volcano-operator/.helmignore
@@ -1,0 +1,22 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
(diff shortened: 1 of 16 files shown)


k82cn:
@SrinivasChilveri , can we use volcano charts from https://github.com/volcano-sh/charts ? So it'll save Kai's review effort when there's any changes in future :)

/cc @thandayuthapani

SrinivasChilveri:
@k82cn , yeah i do agree with you. Its okay to use charts from any where to deploy the volcano in k8s. but i saw in kubeflow/arena all other operators has  respective deployment yaml files to deploy in k8s.

I think @wsxiaozhang  @cheyang  @denverdino  can  take decision

PR #194, opened 2019-07-05
Easily address the GOAPTH in local devlopment enviroment.

xauthulei:
It will be more useful to a go newer in their local.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/194)

PR #195, opened 2019-07-08
Easily address the GOAPTH in local for other README versions

xauthulei:
To address the review comment in PR #194  and keep same in other README versions when referenced  the `GOAPTH`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/195)


xauthulei:
Thanks @cheyang

PR #206, opened 2019-07-12
upgrade arena to support TFJobv1

wsxiaozhang:
- support only one stable TFJob release
- upgrade tfjob crd api to v1
- move tf-operator dependencies from vendor to pkg
- upgrade client-go, k8s.io api to 1.11

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/206)


cheyang:
@wsxiaozhang , I think the origin version of tfjob should be changed from `v1beta1` to `v1alpha1`.

wsxiaozhang:
> @wsxiaozhang , I think the origin version of tfjob should be changed from `v1beta1` to `v1alpha1`.

what do you mean origin version of tfjob?  where do you find v1beta1?

cheyang:
> > @wsxiaozhang , I think the origin version of tfjob should be changed from `v1beta1` to `v1alpha1`.
> 
> @cheyang what do you mean origin version of tfjob? where do you find v1beta1?

See https://github.com/wsxiaozhang/arena/blob/tfjobv1/kubernetes-artifacts/tf-operator/tf-operator.yaml#L51

wsxiaozhang:
> > > @wsxiaozhang , I think the origin version of tfjob should be changed from `v1beta1` to `v1alpha1`.
> > 
> > 
> > @cheyang what do you mean origin version of tfjob? where do you find v1beta1?
> 
> See https://github.com/wsxiaozhang/arena/blob/tfjobv1/kubernetes-artifacts/tf-operator/tf-operator.yaml#L51

@cheyang the yaml is from kubeflow's installation artifacts for tfoperator. the "versions" means which CRD API versions can be supported by this operator. since we upgrade tfoperator to v1, it supports api versions from v1beta2 to v1, and v1 is the default.

wsxiaozhang:
update tfjob CRD supported API versions to V1 and V1Alpha2. v1alpha2 is used to support list existing tfjobs submitted by previous arena tool

Issue #158, opened 2019-05-08
evaluator need Replicas parameter


asdfsx:
When create a tfjob that have `evaluator`, tf-operator will check `Replicas` parameter of `evaluator`. If `Replicas` doesn't exist, it will cause runtime problem. And the tf-operator will crashed.   
Right now the tfjob.yaml doesn't have such parameter.

https://github.com/kubeflow/arena/blob/f3a359432873ef871356a0bb2153249ee0e84ed9/charts/tfjob/templates/tfjob.yaml#L728-L730

And following is the validation code in `tf-operator`

https://github.com/kubeflow/tf-operator/blob/3bf0ac83924aeea57df3e7c50bc82fc3f5546844/pkg/apis/tensorflow/validation/validation.go#L95-L97

cheyang:
@asdfsx , would you like to help on this? Thanks.

asdfsx:
@cheyang fix the issue by pr #257

PR #257, opened 2019-09-04
add replicas in chief & evaluator

asdfsx:
add replicas in chief & evaluator

Fixes issue #158 

According to [multi_worker_util.py#L49-L60](https://github.com/tensorflow/tensorflow/blob/1916f19bd1a792d3ba5e9e0666fca5e6374cf298/tensorflow/python/distribute/multi_worker_util.py#L49-L60), I think tensorflow only support one chief/evaluator.

So I set 1 to both replicas.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/257)


cheyang:
@asdfsx could you please update the chart version and changelog. Thanks.

PR #271, opened 2019-10-20
Refactor code/job

happy2048:
refactor code for job and serving job

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/271)

### DIFF
--- a/cmd/arena/commands/get_advanced.go
+++ b/cmd/arena/commands/get_advanced.go
@@ -65,39 +65,45 @@ func isTrainingConfigExist(name, trainingType, namespace string) bool {
 /**
 * BuildTrainingJobInfo returns types.TrainingJobInfo
  */
-func BuildJobInfo(job TrainingJob) *types.JobInfo {
+func BuildJobInfo(job TrainingJob) *types.TrainingJobPrinterInfo {
 
 	tensorboard, err := tensorboardURL(job.Name(), job.ChiefPod().Namespace)
 	if tensorboard == "" || err != nil {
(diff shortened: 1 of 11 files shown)


pkg/util/k8sutil/k8sutil.go
```diff
@@ -0,0 +1,127 @@
+package k8sutil
```
cheyang: I suggest we also named it to `kube` like Helm. https://github.com/helm/helm/tree/master/pkg/kube

pkg/util/common/file.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License
+
+package common
+
+import (
+	"os"
+)
+
+func CheckFileExist(filename string) bool {
```
cheyang: Why do we need this function?

cmd/arena/commands/get_advanced.go
```diff
@@ -65,39 +65,45 @@ func isTrainingConfigExist(name, trainingType, namespace string) bool {
 /**
 * BuildTrainingJobInfo returns types.TrainingJobInfo
  */
-func BuildJobInfo(job TrainingJob) *types.JobInfo {
+func BuildJobInfo(job TrainingJob) *types.TrainingJobPrinterInfo {
```
cheyang: Could you make the comment and method name consistent? Why using TrainingJobPrinterInfo to replace TrainingJob?

PR #277, opened 2019-11-26
volcano trainer isSupported for the specific job

ocherfas:
Currently Volcano is supported function return true if any volcano jobs exists in the system.
This created problem with list commands.

It should return true for the specific job name.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/277)


ocherfas:
@googlebot I signed it!

cheyang:
Thanks for your contribution! @ocherfas

PR #281, opened 2019-12-02
horovod and standalone gets the only the jobs they created

ocherfas:
horovod and standalone trainer should list only their own job resources.

We use our custom trainer that utilize the Job resource of kubernetes.
**What is happening**
1. Create a new job with our custom trainer
2. Get job using job name without specific `--type` flag
3. Gets an error "There are more than 1 training jobs with the same name ...", Altough only job job exists in the system
**What should happend**
1. Create a new job with our custom trainer.
2. Get job using job name without specific `--type` flag
3. Our job gets displayed

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/281)


cheyang:
Thanks for your contribution!

PR #294, opened 2020-02-05
fix incorrect cpu resource variable, should be psCPU

meibenjin:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/arena/294)


cheyang:
Thanks very much for the contribution! @meibenjin

PR #309, opened 2020-03-31
Change deprecated flag  --syncMode and --syncSource

ddbmh:
Change deprecated flag  --syncMode and --syncSource of file 1-tfjob-standalone.md


ddbmh:
@googlebot I signed it

PR #313, opened 2020-05-08
support adding gpu resource to ps role pods for submitting tf job

happy2048:
1.support assigning gpu resource to ps role pods for submitting tf jobs
2.if ps role pods want to request gpus,use option '--ps-gpus' to achieve it

### DIFF
--- a/charts/tfjob/CHANGELOG.md
+++ b/charts/tfjob/CHANGELOG.md
@@ -120,3 +120,7 @@
 ### 0.28.0
 
 * PS annotations should be a attribute of metadata yaml node.
+
+### 0.29.0
+
+* support assgining gpu resources for PS when submitting tfjobs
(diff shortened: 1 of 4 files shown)


charts/tfjob/CHANGELOG.md
```diff
@@ -120,3 +120,7 @@
 ### 0.28.0
 
 * PS annotations should be a attribute of metadata yaml node.
+
+### 0.29.0
+
+* support assgining gpu resources for ps role pods when submitting tfjobs
```
cheyang, line 126: ps role pods -> PS

charts/tfjob/templates/tfjob.yaml
```diff
@@ -228,6 +228,9 @@ spec:
               {{- end }}
               resources:             
                 requests:
+                  {{- if .Values.psGPU }}
```
cheyang, line 231: why not  using {{- if gt (int $gpuCount) 0}}  ?

PR #318, opened 2020-06-01
fix user guide about submit tfjob

jiaqianjing:


jiaqianjing:
@googlebot I signed it!

PR #329, opened 2020-06-28
add tips for dataset that pvc name is not support repeated #29005224

jiaqianjing:


pkg/util/volume.go
```diff
@@ -28,6 +28,7 @@ const restrictedNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.-]`
 var restrictedNamePattern = regexp.MustCompile(`^` + restrictedNameChars + `+$`)
 
 func ValidateDatasets(dataset []string) (err error) {
+	tempPVCList := make(map[string]string)
```
cheyang, line 31: please use map[string]bool

jiaqianjing: ok

PR #355, opened 2020-08-14
disabled launcher of mpi job to mount pvc

happy2048:
1.disable laucher of mpi job to mount pvc


kubernetes-artifacts/mpi-operator/mpi-operator.yaml
```diff
@@ -170,4 +170,4 @@ spec:
           "-v=5"
         ]
         imagePullPolicy: Always
----
```
cheyang, line 173: return new line

PR #362, opened 2020-09-25
update et-controller image version

xiaozhouX:
update et controller image tag


xiaozhouX:
@googlebot I fixed it.

PR #455, opened 2021-02-03
Feature/update et operator crd

xiaozhouX:
Update et operator crd, use `apiextensions.k8s.io/v1beta1` crd version.


xiaozhouX:
@googlebot I consent

PR #516, opened 2021-03-29
chore: do not track __pycache__, dist, build, egg-info folder

haijohn:
remove \_\_pycache\_\_, dist, build, egg-info folder, and update .gitignore


haijohn:
@googlebot I signed it!

PR #569, opened 2021-05-28
Fix cmd typo in installation/binary.md

uzuku:
Fix cmd typo.


uzuku:
@googlebot I signed it!

PR #575, opened 2021-06-04
fix install error on mac

goodoid:
if we rm and cp arena-kubectl for each installation, mac will notify "open app from unknown developer" continuously.


goodoid:
> @googlebot I signed it!

goodoid:
@googlebot I signed it!

PR #617, opened 2021-07-23
add annotation in tf job level

denkensk:
add the annotation in tf job level. not only in pod level.


charts/tfjob/templates/tfjob.yaml
```diff
@@ -14,6 +14,10 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
     createdBy: "TFJob"
+  annotations:
```
cheyang, line 17: Please also update chart version and change log

cheyang (changes requested):

PR #619, opened 2021-07-26
support annotations/nodeSelector/tolerations in tensorflow serving an…

heluocs:
support annotations/nodeSelector/tolerations in arena serve tfserving and triton


cheyang (changes requested):
Please also update chart version and change log

PR #825, opened 2022-07-27
add pod info for serving deployment

tingshua-yts:
add ARENA_NODE_NAME /ARENA_POD_NAMESPACE/ ARENA_POD_NAME / ARENA_POD_IP for triton/custom/tfserving deployment

### DIFF
--- a/charts/tfserving/templates/deployment.yaml
+++ b/charts/tfserving/templates/deployment.yaml
@@ -71,7 +71,7 @@ spec:
       initContainers:
 
       containers:
-        - name: serving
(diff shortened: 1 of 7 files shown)


charts/tfserving/templates/deployment.yaml
```diff
@@ -71,7 +71,7 @@ spec:
       initContainers:
 
       containers:
-        - name: serving
+        - name: serving
```
cheyang, line 74: Please remove the space. thanks.

PR #826, opened 2022-07-28
Add support of gpu-core in top-node command.

AlanFokCo:


pkg/topnode/gpushare.go
```diff
@@ -38,6 +44,17 @@ GPU Summary:
...
 
+var gpushareWithGpuCoreSummary = `
+GPU Summary:
+  Total GPUs:           %v
+  Allocated GPUs:       %v
+  Unhealthy GPUs:       %v
+  Total GPU Memory:     %.1f GiB
+  Allocated GPU Memory: %.1f GiB
+  Used GPU Memory:      %d
+  Total GPU Core:		%d
```
happy2048, line 55: need to be aligned with other lines

PR #832, opened 2022-08-01
add pod info

tingshua-yts:


cheyang (changes requested):
Please update chart version and change log

PR #835, opened 2022-08-04
add evaluator and tensorboard to pod group

KunWuLuan:
fix(tfjob):add evaluator and tensorboard to pod group


happy2048 (commented):
please add chart README.md and update chart version

PR #848, opened 2022-10-12
modify restful-serving to http-serving of serve deployment service port name

tingshua-yts:
Since Istio require the prefix of  service port name must be protocol, e.g. http-serving, we change arena serve deployment service port name to support this.

### DIFF
--- a/charts/custom-serving/templates/service.yaml
+++ b/charts/custom-serving/templates/service.yaml
@@ -28,7 +28,7 @@ spec:
       targetPort: {{ .Values.restApiPort }}
     {{- end }}
     {{- if ne (int .Values.metricsPort) 0 }}
-    - name: metrics
+    - name: http-metrics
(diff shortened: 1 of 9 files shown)


charts/custom-serving/templates/service.yaml
```diff
@@ -23,12 +23,12 @@ spec:
...
     {{- end }}
     {{- if ne (int .Values.restApiPort) 0 }}
-    - name: restful-serving
+    - name: http-serving
       port: {{ .Values.restApiPort }}
       targetPort: {{ .Values.restApiPort }}
     {{- end }}
     {{- if ne (int .Values.metricsPort) 0 }}
-    - name: metrics
+    - name: http-metrics
```
cheyang, line 31: Is http-metrics required by the client?

cheyang (approved):

PR #879, opened 2022-11-29
Add display of startTime and endTime on getting and topping job.

AlanFokCo:


AlanFokCo:
A Running Job (arena top and arena get)
image
A Succeed Job (arena top and arena get)
image

PR #907, opened 2023-02-17
Fix model serve args bug.

AlanFokCo:
When I use 'arena serve tf' to submit a TfServing, if I want to use '--gpumemory', arena will fill the server args with
'--per_process_gpu_memory_fraction=$(awk 'BEGIN{printf"%.2f",'$ALIYUN_COM_GPU_MEM_CONTAINER'/'$ALIYUN_COM_GPU_MEM_DEV'}')'.
So it will have a panic:
panic: runtime error: index out of range [1] with length 1
And if other unexpected parameters appear, this problem will appear again.

### DIFF
--- a/pkg/model/model.go
+++ b/pkg/model/model.go
@@ -263,7 +263,6 @@ func (m *modelJob) Params() map[string]string {
 	for _, value := range arr {
 		if strings.HasPrefix(value, "--") {
 			kv := strings.Split(value, "=")
-			params[kv[0]] = kv[1]
 			params[fmt.Sprintf("--%s", kv[0])] = value[len(kv[0])+1:]
 		}
 	}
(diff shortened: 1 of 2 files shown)


AlanFokCo:
When I use 'arena serve tf' to submit a TfServing, if I want to use '--gpumemory', arena will fill the server args with
`'--per_process_gpu_memory_fraction=$(awk 'BEGIN{printf "%.2f",'$ALIYUN_COM_GPU_MEM_CONTAINER'/'$ALIYUN_COM_GPU_MEM_DEV'}')'`. 
So it will have a panic:
`panic: runtime error: index out of range [1] with length 1`
And if other unexpected parameters appear, this problem will appear again.

PR #930, opened 2023-03-03
Support TFJob/PytorchJob ActiveDeadlineSeconds

Syulin7:
tf-operator and pytorch-operator already support ActiveDeadlineSeconds.

1. support it in arena submit job.
2. pytorchjob default CleanPodPolicy change to Running

### DIFF
--- a/pkg/argsbuilder/submit_pytorchjob.go
+++ b/pkg/argsbuilder/submit_pytorchjob.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//       http://www.apache.org/licenses/LICENSE-2.0
+//	http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
@@ -15,13 +15,15 @@ package argsbuilder
 
 import (
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"reflect"
 	"strings"
+	"time"
 
-	"github.com/kubeflow/arena/pkg/apis/types"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/kubeflow/arena/pkg/apis/types"
 )
 
 type SubmitPytorchJobArgsBuilder struct {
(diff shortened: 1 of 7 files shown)


pkg/argsbuilder/submit_pytorchjob.go
```diff
@@ -72,6 +72,7 @@ func (s *SubmitPytorchJobArgsBuilder) AddCommandFlags(command *cobra.Command) {
 	command.Flags().StringVar(&s.args.CleanPodPolicy, "clean-task-policy", "None", "How to clean tasks after Training is done, support None, Running, All.")
 	command.Flags().StringVar(&s.args.Cpu, "cpu", "", "the cpu resource to use for the training, like 1 for 1 core.")
 	command.Flags().StringVar(&s.args.Memory, "memory", "", "the memory resource to use for the training, like 1Gi.")
+	command.Flags().Int64Var(&s.args.ActiveDeadlineSeconds, "active-deadline-seconds", 0, "Specifies the duration (in seconds) since startTime during which the job can remain active before it is terminated")
```
cheyang, line 75: I think  the default value 0 for active-deadline-seconds will mislead the user. How about not setting default value? If it's not setting,  There is no ActiveDeadlineSeconds.

cheyang (changes requested):

pkg/argsbuilder/submit_pytorchjob.go
Syulin7: Done, If it's not setting, There is no ActiveDeadlineSeconds in job's CR.

pkg/argsbuilder/submit_tfjob.go
```diff
@@ -130,6 +133,8 @@ func (s *SubmitTFJobArgsBuilder) AddCommandFlags(command *cobra.Command) {
 	command.Flags().MarkDeprecated("cleanTaskPolicy", "please use --clean-task-policy instead")
 	command.Flags().StringVar(&s.args.CleanPodPolicy, "clean-task-policy", "Running", "How to clean tasks after Training is done, support Running, None and All.")
 
+	command.Flags().DurationVar(&activeDeadline, "active-deadline", activeDeadline, "Specifies the duration since startTime during which the job can remain active before it is terminated(e.g. '5s', '1m', '2h22m').")
```
cheyang, line 136: I suggest using the name `running-timeout` instead of `active-deadline`.

Syulin7: Done

PR #945, opened 2023-03-10
Add imagePullSecret and shareMemory for arena serve.

AlanFokCo:
Add new args for arena serve: --image-pull-secret and --share-memory.


AlanFokCo:
Add usage information for extend-command.

PR #947, opened 2023-03-30
Add ownerReference for configmap and tensorboard

Syulin7:
add ownerReference  for configmap and tensorboard:
1. patch configmap ownerReference.
2. patch deployment/tensorboard ownerReference.
3. add label in service/tensorboard, and job operator will add ownerReferences.
see: https://github.com/AliyunContainerService/tf-operator/blob/b81dfd7a95c21f0f9bf6e8030d591edd9415c1e9/pkg/control/service_ref_manager.go#L140

and fix etjob gitImage which didn't exist before.

### DIFF
--- a/pkg/util/kubectl/kubectl.go
+++ b/pkg/util/kubectl/kubectl.go
@@ -16,18 +16,21 @@ package kubectl
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
-	"github.com/kubeflow/arena/pkg/apis/config"
 	"io/ioutil"
-	v1 "k8s.io/api/apps/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
 	"os"
 	"os/exec"
 	"strings"
 	"sync"
 
 	log "github.com/sirupsen/logrus"
+	v1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/kubeflow/arena/pkg/apis/config"
 )
 
 var kubectlCmd = []string{"arena-kubectl"}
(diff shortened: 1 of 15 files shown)


Syulin7:
@cheyang @happy2048 PTAL, Thanks.

pkg/util/kubectl/kubectl.go
```diff
@@ -354,3 +358,75 @@ func UpdateDeployment(deploy *v1.Deployment) error {
...
+	}
+	resources := strings.Split(string(data), "\n")
+	// add configmap
+	configmapName := fmt.Sprintf("%v-%v", name, trainingType)
+	resources = append(resources, "configmap/"+configmapName)
+
+	var wg sync.WaitGroup
+	locker := new(sync.RWMutex)
+	for _, r := range resources {
+		resource := r
```
cheyang, line 400: I suggest using `for` loop rather than ` goroutine concurrency`. And is there any way to recover if the API Server is not available in short time?

pkg/util/kubectl/kubectl.go
```diff
@@ -354,3 +358,75 @@ func UpdateDeployment(deploy *v1.Deployment) error {
...
+		// skip training job.
+		// skip tensorboard service: job operator will add ownerReferences.
+		if strings.Contains(resource, "tfjob.kubeflow.org/") ||
+			strings.Contains(resource, "pytorchjob.kubeflow.org/") ||
```
cheyang, line 404: When handling resource objects, the current code uses the strings.Contains function to determine whether to perform the JSON patch operation based on the resource name. This approach may not be accurate. Consider using a more accurate way to determine the resource type, such as using the official Kubernetes API or resource label information.

pkg/util/kubectl/kubectl.go
Syulin7: Done

pkg/util/kubectl/kubectl.go
```diff
@@ -354,3 +358,75 @@ func UpdateDeployment(deploy *v1.Deployment) error {
...
+	// get training job
+	args := []string{binary, "get", trainingType, name, "--namespace", namespace, "-o json"}
+	cmd := exec.Command("bash", "-c", strings.Join(args, " "))
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		errs = append(errs, string(out))
+	}
+
+	obj := &unstructured.Unstructured{}
+	err = json.Unmarshal(out, obj)
```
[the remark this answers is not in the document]
Syulin7: Done

PR #953, opened 2023-04-18
Move policy v1beta1 to v1.

AlanFokCo:
The api policy/v1beta1 will be deprecated in versions greater than kubernetes 1.26.


cheyang (approved):
/lgtm
/approve

PR #954, opened 2023-05-11
Fix evaluatejob job yaml in charts.

AlanFokCo:
The rendering of imagePullSecrets in the chart does not conform to the format.


cheyang (commented):
/lgtm
/approve

PR #960, opened 2023-06-13
Support new training type deepspeed.

Syulin7:
Support new training type deepspeedjob. for more information，usage:
arena submit deepspeedjob --help or arena submit dp --help

### DIFF
--- a/charts/etjob/CHANGELOG.md
+++ b/charts/etjob/CHANGELOG.md
@@ -14,3 +14,6 @@
 
 * update gitImage
 
+### 0.3.2
+
+* Support new training type deepspeed.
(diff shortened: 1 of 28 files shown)


cheyang (approved):
/lgtm
/approve

panpan0000:
Can you provide DOC update as well ?

Cannot catch what is the difference between deepspeed job than other type of job.

Syulin7:
@panpan0000  Sorry for late reply.

I will update the documentation later. 

deepspeedjob and mpijob are similar in that they follow the launcher-worker pattern. The launcher sends commands to the workers using pdsh and is implemented based on et-operator. It automatically creates the configuration files needed by deepspeed, such as /job/hostfile.

PR #962, opened 2023-06-20
[FIX] update serve duplicate create env and toleration

Syulin7:
when update serve and set same env or toleration，it will create duplicate env or toleration.
```
        env:
        - name: yulin
          value: qwe
        - name: yulin
          value: qwe
        - name: yulin
          value: qwe
```


cheyang (approved):
/lgtm
/approve

PR #964, opened 2023-06-28
fix --data-dir is not taking effect in custom-serving.

Syulin7:
fix --data-dir is not taking effect in custom-serving.


cheyang (approved):
/lgtm
/approve

PR #969, opened 2023-08-01
Update arena version to 0.9.10

Syulin7:
- Fix --data-dir is not taking effect in custom-serving.
- Fix the prompt content when submitting serve job.
- Default delete secret permissions in et-operator.
- Enable create secret for deepspeedjob, etjob.


cheyang (changes requested):
Will you update arena version 0.9.10 accordingly?

cheyang (approved):
/lgtm
/approve

Issue #966, opened 2023-07-12
The cleanPolicy field of PyTorchJob is incorrectly placed.


lvxian-kjl:
The cleanPolicy, activeDeadlineSeconds, and ttlSecondsAfterFinished within spec should be under spec.runPolicy, not directly under spec.

panpan0000:
+1

I just see the same problem.


v0.9.0+447b534 still have this problem.

panpan0000:
Hi, @cheyang , considering the incompatibility between the kubeflow trainer-operator CRD with arena code , what's your though ?

panpan0000:
If no objection, I'm doing the fix now

Syulin7:
@lvxian-kjl @panpan0000   In previous tf-operator and pytorch-operator, the cleanPolicy was under spec, but in the training-operator CRD, it has been moved to spec.RunPolicy, which is a breaking change. Considering compatibility with users who are currently using it, there has been no direct switch to the training-operator CRD.

Syulin7:
> If no objection, I'm doing the fix now

Thank you for your PR, but it would impact existing users. We are also considering how to maintain compatibility with the training-operator CRD, We are also considering how to make it compatible with the training-operator CRD, such as adding a new template in the arena charts, and making a distinction when installing arena.

panpan0000:
Thank you @Syulin7 , but people's intuitions are "arena will be the CLI tool to manage the jobs , after installing kubeflow and its training-operator"
But the current arena will never submit any job for up to date kubeflow standard training-operator and CRDs.
So every new user will be block here and has to give up arena. I'm afraid that's not the good way out for community.

May I suggest that 
(1) create a maintenance branch to support those OLD users with old kubeflow CRD, especially for those users in AliYun. with NOTES in README to tell the max compatbile kubeflow version, to avoid confusion  .
(2) the main branch just supports latest kubeflow CRD, and add a NOTE in README,  to tell people the minimal kubeflow version for this branch.

Syulin7:
@panpan0000 Thank you for your suggestion, I agree. Supporting the training-operator CRD and maintaining backward compatibility are both very important.

Creating a new branch is a solution, but it may add complexity and cause confusion. I am trying to achieve compatibility with the training-operator CRD by identifying the CRD and using different templates when submitting jobs. In summary, our goal is to support the latest training-operator CRD.

panpan0000:
[...]

PR #1015, opened 2023-12-18
Fix: #966, correct cleanPolicy position

panpan0000:


Syulin7:
Thank you for your PR, but it's a breaking change that will prevent installations of the older version of arena from being upgraded to the latest version.

panpan0000:
let's discuss in the thread of original issue

> 
> Thank you @Syulin7 , but people's intuitions are "arena will be the CLI tool to manage the jobs , after installing kubeflow and its training-operator"
> But the current arena will never submit any job for up to date kubeflow standard training-operator and CRDs.
> So every new user will be blocked here and has to give up arena. I'm afraid that's not the good way out for community.
> 
> May I suggest that :
> (1) create a maintenance branch to support those OLD users with old kubeflow CRD, especially for those users in AliYun. with NOTES in README to tell the max compatbile kubeflow version, to avoid confusion .
> (2) the main branch just supports latest kubeflow CRD, and add a NOTE in README, to tell people the minimal kubeflow version for this branch.

PR #1036, opened 2024-02-04
update tritonserver base image to nvcr.io/nvidia/tritonserver:24.01-py3

gujingit:
# Feature
update tritonserver base image to nvcr.io/nvidia/tritonserver:24.01-py3


Syulin7:
Thanks @gujingit 
/lgtm
/approve

PR #1109, opened 2024-07-22
Fix incorrect TensorBoard images.

Syulin7:
Fix incorrect TensorBoard images.


ChenYi015 (approved):
Thanks for the contribution!

PR #1113, opened 2024-07-31
Remove docker dependency

Syulin7:
Remove docker dependency


ChenYi015 (approved):
Thanks for the contribution!

PR #1130, opened 2024-09-12
Increase RSA key bit size from 1024 to 2048

ChenYi015:

Increase RSA key bit size from 1024 to 2048


cheyang (approved):
/lgtm
/approve

PR #1207, opened 2024-11-16
refactor: use helm lib instead of helm binary

ChenYi015:
## Purpose of this PR

**Proposed changes:**

- Use helm lib instead of helm binary, so that arena itself will not depend on `arena-helm` anymore.

## Change Category

- [ ] Documentation update

### Additional Context

- `helm` will still be packaged into arena-installer, as the install.sh will rename it to `arena-helm`.
- `arena-helm` will only be used to install arena artifacts to k8s e.g. CRDs and various operators.


pkg/training/delete.go
```diff
@@ -30,13 +29,7 @@ func DeleteTrainingJob(jobName, namespace string, jobType types.TrainingJobType)
 	if jobType == types.UnknownTrainingJob {
 		return fmt.Errorf("Unsupport job type,arena only supports: [%v]", utils.GetSupportTrainingJobTypesInfo())
 	}
-	// 1. Handle legacy training job
-	err := helm.DeleteRelease(jobName)
-	if err == nil {
-		log.Infof("Delete the job %s successfully.", jobName)
-		return nil
-	}
-	log.Debugf("%s wasn't deleted by helm due to %v", jobName, err)
```
ChenYi015, line 39: Remove this for `helm.DeleteRelease` is out of date and will always return non-nil error.

cheyang:
Thanks for great contributions. @ChenYi015 I suggest building robust e2e testing first before  refactoring the current implementation.

cheyang:
I suggest avoiding breaking changes that lack forward compatibility.  For example, we can introduce `helm-binray` for forward compatibility and `helm` for helm library, and use environment variable to switch the way of using helm.

ChenYi015:
> I suggest avoiding breaking changes that lack forward compatibility. For example, we can introduce `helm-binray` for forward compatibility and `helm` for helm library, and use environment variable to switch the way of using helm.

Have added a new persistent flag `--helm-binary` for forward compatibility. If `--helm-binary` is specified, `arena-helm` binary will be used to render templates.

PR #1228, opened 2024-12-18
feat: add support for torchrun

ChenYi015:
## Purpose of this PR

**Proposed changes:**

- Add support for torchrun

## Change Category

- [ ] Documentation update

### DIFF
--- a/pkg/apis/types/submit_pytorchjob.go
+++ b/pkg/apis/types/submit_pytorchjob.go
@@ -45,4 +45,8 @@ type SubmitPyTorchJobArgs struct {
 
 	// ShareMemory Specifies the shared memory size
 	ShareMemory string `yaml:"shareMemory"`
+
+	// Number of workers per node, supported values: [auto, cpu, gpu, int].
+	// For more, https://github.com/pytorch/pytorch/blob/26f7f470df64d90e092081e39507e4ac751f55d6/torch/distributed/run.py#L629-L658.
+	NprocPerNode string `yaml:"nprocPerNode,omitempty"`
 }
--- a/pkg/argsbuilder/submit_pytorchjob.go
+++ b/pkg/argsbuilder/submit_pytorchjob.go
@@ -16,6 +16,7 @@ package argsbuilder
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -83,6 +84,7 @@ func (s *SubmitPytorchJobArgsBuilder) AddCommandFlags(command *cobra.Command) {
 	command.Flags().DurationVar(&runningTimeout, "running-timeout", runningTimeout, "Specifies the duration since startTime during which the job can remain active before it is terminated(e.g. '5s', '1m', '2h22m').")
 	command.Flags().DurationVar(&ttlAfterFinished, "ttl-after-finished", ttlAfterFinished, "Defines the TTL for cleaning up finished PytorchJobs(e.g. '5s', '1m', '2h22m'). Defaults to infinite.")
 	command.Flags().StringVar(&s.args.ShareMemory, "share-memory", "2Gi", "the shared memory of each replica to run the job, default 2Gi.")
+	command.Flags().StringVar(&s.args.NprocPerNode, "nproc-per-node", "", "The number of processes per node, available values are \"auto\", \"cpu\", \"gpu\" and a number (e.g. 4).")
 
 	s.AddArgValue("running-timeout", &runningTimeout).
 		AddArgValue("ttl-after-finished", &ttlAfterFinished)
@@ -110,7 +112,7 @@ func (s *SubmitPytorchJobArgsBuilder) Build() error {
 	if err := s.check(); err != nil {
 		return err
 	}
-	if err := s.addMasterAddr(); err != nil {
+	if err := s.addEnv(); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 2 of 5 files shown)


pkg/apis/types/submit_pytorchjob.go
```diff
@@ -45,4 +45,8 @@ type SubmitPyTorchJobArgs struct {
 
 	// ShareMemory Specifies the shared memory size
 	ShareMemory string `yaml:"shareMemory"`
+
+	// Number of workers per node, supported values: [auto, cpu, gpu, int].
+	// For more, https://github.com/pytorch/pytorch/blob/26f7f470df64d90e092081e39507e4ac751f55d6/torch/distributed/run.py#L629-L658.
+	NprocPerNode string `yaml:"nprocPerNode,omitempty"`
```
cheyang, line 51: I suggest using  numbers for NprocPerNode is a sensible choice given its literal meaning, as it explicitly specifies the number of processes per node.  And by default it's 1 to make consistency with previous version.

pkg/argsbuilder/submit_pytorchjob.go
```diff
@@ -173,12 +174,17 @@ func (s *SubmitPytorchJobArgsBuilder) check() error {
...
-func (s *SubmitPytorchJobArgsBuilder) addMasterAddr() error {
+func (s *SubmitPytorchJobArgsBuilder) addEnv() error {
 	if s.args.EnableRDMA {
 		if s.args.Envs == nil {
 			s.args.Envs = map[string]string{}
 		}
 		s.args.Envs["MASTER_ADDR"] = fmt.Sprintf("%v-master-0", s.args.Name)
 	}
+
+	if s.args.NprocPerNode != "" {
```
cheyang, line 185: Do we need to determine if s.args.Envs is empty pointer here?

cheyang (changes requested):

pkg/argsbuilder/submit_pytorchjob.go
ChenYi015: Done.

pkg/apis/types/submit_pytorchjob.go
ChenYi015: I think we should keep the `--nproc-per-node` the same way as `torchrun`:

```bash
$ torchrun --help
...
options:
  -h, --help            show this help message and exit
  --nnodes NNODES       Number of nodes, or the range of nodes in form <minimum_nodes>:<maximum_nodes>.
  --nproc-per-node NPROC_PER_NODE, --nproc_per_node NPROC_PER_NODE
                        Number of workers per node; supported values: [auto, cpu, gpu, int].
...
```

The available values are `auto`, `cpu`, `gpu` and an integer. If `--nproc-per-node` is unset, then `1` will be used, ref: https://github.com/pytorch/pytorch/blob/a8d6afb511a69687bbb2b7e88a3cf67917e1697e/torch/distributed/run.py#L433-L440.

For `arena`, if `--nproc-per-node` is unset, then we will not ingest env `PET_NPROC_PER_NODE` to the pod template. Otherwise, we will check whether the value specified by `--nproc-per-node` is valid and append it to the pod template envs.

pkg/argsbuilder/submit_pytorchjob.go
```diff
@@ -170,15 +172,36 @@ func (s *SubmitPytorchJobArgsBuilder) check() error {
...
+	}
+
 	if s.args.EnableRDMA {
-		if s.args.Envs == nil {
-			s.args.Envs = map[string]string{}
-		}
 		s.args.Envs["MASTER_ADDR"] = fmt.Sprintf("%v-master-0", s.args.Name)
 	}
+
+	if s.args.NprocPerNode != "" {
```
cheyang, line 202: If the user set --env=PET_NPROC_PER_NODE=1. How to handle this logic?

pkg/argsbuilder/submit_pytorchjob.go
```diff
@@ -83,6 +84,7 @@ func (s *SubmitPytorchJobArgsBuilder) AddCommandFlags(command *cobra.Command) {
 	command.Flags().DurationVar(&runningTimeout, "running-timeout", runningTimeout, "Specifies the duration since startTime during which the job can remain active before it is terminated(e.g. '5s', '1m', '2h22m').")
 	command.Flags().DurationVar(&ttlAfterFinished, "ttl-after-finished", ttlAfterFinished, "Defines the TTL for cleaning up finished PytorchJobs(e.g. '5s', '1m', '2h22m'). Defaults to infinite.")
 	command.Flags().StringVar(&s.args.ShareMemory, "share-memory", "2Gi", "the shared memory of each replica to run the job, default 2Gi.")
+	command.Flags().StringVar(&s.args.NprocPerNode, "nproc-per-node", "", "The number of processes per node, available values are \"auto\", \"cpu\", \"gpu\" and a number (e.g. 4).")
```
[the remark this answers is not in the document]
ChenYi015: The default value is empty and arena will not ingest env `PET_NPROC_PER_NODE` to the pod template.

pkg/argsbuilder/submit_pytorchjob.go
ChenYi015: > If the user set --env=PET_NPROC_PER_NODE=1. How to handle this logic?

If the user set both `--nproc-per-node=2` and `--env=PET_NPROC_PER_NODE=1`, then there will be two environment variables named `PET_NPROC_PER_NODE` with values `1` and `2` respectively. Otherwise, there will only be one env `PET_NPROC_PER_NODE=1`.

cheyang (approved):
/lgtm
/approve

PR #1289, opened 2025-03-20
fix: job status displays incorrectly

ChenYi015:
## Purpose of this PR

**Proposed changes:**

- fix: job status displays incorrectly

## Change Category

- [ ] Documentation update

### DIFF
--- a/pkg/training/trainer_tensorflow.go
+++ b/pkg/training/trainer_tensorflow.go
@@ -93,10 +93,15 @@ func (tj *TensorFlowJob) GetLabels() map[string]string {
 
 // GetStatus returns the status of the Job i.e. PENDING, QUEUING, RUNNING, SUCCEEDED and FAILED.
 func (tj *TensorFlowJob) GetStatus() string {
+	status := "PENDING"
+	defer log.Debugf("Get status of TFJob %s: %s", tj.tfjob.Name, status)
+
 	if tj.tfjob.Name == "" {
-		return "PENDING"
+		return status
 	}
-	return getStatus(tj.tfjob.Status)
+
+	status = getStatus(tj.tfjob.Status)
+	return status
 }
 
 // StartTime returns the start time
@@ -436,7 +441,8 @@ func getStatus(status commonv1.JobStatus) string {
 	} else {
 		s = "Pending"
 	}
-	return strings.ToUpper(s)
+	s = strings.ToUpper(s)
+	return s
 }
 
 // hasCondition checks if the given job status has the condition type.
--- a/pkg/training/trainer_tensorflow.go
+++ b/pkg/training/trainer_tensorflow.go
@@ -19,23 +19,20 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
-	"github.com/kubeflow/arena/pkg/operators/tf-operator/client/clientset/versioned"
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/training/trainer_tensorflow.go
```diff
@@ -433,17 +422,31 @@ func makeJobStatusSortedByTime(conditions []commonv1.JobCondition) []commonv1.Jo
...
-func checkStatus(status commonv1.JobStatus) commonv1.JobConditionType {
-	t := commonv1.JobConditionType("Pending")
+// getStatus returns the status of the Job i.e. PENDING, QUEUING, RUNNING, SUCCEEDED and FAILED.
+func getStatus(status commonv1.JobStatus) string {
+	var s string
+	if hasCondition(status, commonv1.JobSucceeded) {
+		s = string(commonv1.JobSucceeded)
+	} else if hasCondition(status, commonv1.JobFailed) {
+		s = string(commonv1.JobFailed)
+	} else if hasCondition(status, commonv1.JobConditionType("Queuing")) {
```
cheyang, line 432: I think If a job meets multiple conditions (e.g., both Queuing and Running), the order in getStatus() may return an incorrect status.

cheyang (changes requested):

pkg/training/trainer_tensorflow.go
```diff
@@ -433,17 +422,31 @@ func makeJobStatusSortedByTime(conditions []commonv1.JobCondition) []commonv1.Jo
 	return []commonv1.JobCondition(newConditions)
 }
 
-func checkStatus(status commonv1.JobStatus) commonv1.JobConditionType {
-	t := commonv1.JobConditionType("Pending")
+// getStatus returns the status of the Job i.e. PENDING, QUEUING, RUNNING, SUCCEEDED and FAILED.
+func getStatus(status commonv1.JobStatus) string {
```
cheyang, line 426: It's better to add unit test here.

pkg/training/trainer_tensorflow.go
```diff
@@ -433,17 +422,31 @@ func makeJobStatusSortedByTime(conditions []commonv1.JobCondition) []commonv1.Jo
...
+	if hasCondition(status, commonv1.JobSucceeded) {
+		s = string(commonv1.JobSucceeded)
+	} else if hasCondition(status, commonv1.JobFailed) {
+		s = string(commonv1.JobFailed)
+	} else if hasCondition(status, commonv1.JobConditionType("Queuing")) {
+		s = "Queuing"
+	} else if hasCondition(status, commonv1.JobRunning) {
+		s = string(commonv1.JobRunning)
+	} else {
+		s = "Pending"
```
cheyang, line 437: I think it's better to log the value if it's not empty string for debugging.

cheyang (changes requested):

pkg/training/trainer_tensorflow.go
ChenYi015: Added.

pkg/training/trainer_tensorflow.go
ChenYi015: Have added a logging statement at `PytorchJob.GetStatus` and `TensorflowJob.GetStatus`.

pkg/training/trainer_tensorflow.go
cheyang: I think if it's possible in logic, we should handle this.

cheyang (approved):
/lgtm
/approve

PR #1296, opened 2025-04-02
feat: add new value for using localtime in cron-operator

ChenYi015:
## Purpose of this PR

**Proposed changes:**

- Mount host localtime into corn-operator pod, so users are able to write a cron schedule with local timezone rather than UTC timezone.

## Change Category

- [ ] Documentation update

### DIFF
--- a/arena-artifacts/values.yaml
+++ b/arena-artifacts/values.yaml
@@ -112,8 +112,8 @@ cron:
   image: acs/cron-operator
   tag: aliyun-0f570ce
   imagePullPolicy: IfNotPresent
-  # -- Whether to use local time in the container.
-  localTime: false
+  # -- Whether to use host timezone in the container.
+  useHostTimezone: false
   resources:
     limits:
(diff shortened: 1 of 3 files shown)


arena-artifacts/values.yaml
```diff
@@ -112,6 +112,8 @@ cron:
   image: acs/cron-operator
   tag: aliyun-0f570ce
   imagePullPolicy: IfNotPresent
+  # -- Whether to use local time in the container.
+  localTime: false
```
cheyang, line 116: I suggest it's better to use flag name  `useHostTimezone`   instead of `localTime`.

cheyang (changes requested):

arena-artifacts/values.yaml
ChenYi015: Done. Thanks for your review.

cheyang (approved):
/lgtm
/approve

PR #1331, opened 2025-06-04
Feat: support separate affinity policy configuration for PS and worke…

Syspretor:
For tfjob, PS and worker typically have different affinity configuration requirements, and an independent configuration flag is needed. This PR provides separate affinity configuration options for PS and worker in tfjob.
```
--ps-affinity-policy
--ps-affinity-constraint
--worker-affinity-policy
--worker-affinity-constraint
```


cheyang (approved):
/lgtm
/approve

PR #1407, opened 2026-01-25
fix: add ownerReferences to resources created by arena

### LINKED ISSUE #1339: Add ownerReferences to resources created by arena
### What you would like to be added?

When one uses Arena to submit a training job with `ttlSecondsAfterFinished` set, the training job will be cleaned up when the TTL is reached, but the associated resources created by Arena will not.

### Why is this needed?

Also clean up the resources created by Arena once the training job is deleted. Otherwise, there will be a large amount of resources (e.g. ConfigMaps) may remain uncleaned, leading to the resource leaks in the cluster.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

FAUST-BENCHOU:
## Purpose of this PR
Close https://github.com/kubeflow/arena/issues/1339

**Proposed changes:**

Call PatchOwnerReferenceWithAppInfoFile to add an ownerReference to all resources, pointing to the training job's CRD.

## Change Category

- [ ] Documentation update

### DIFF
--- a/pkg/util/kubectl/kubectl.go
+++ b/pkg/util/kubectl/kubectl.go
@@ -33,6 +33,7 @@ import (
 	lwsClient "sigs.k8s.io/lws/client-go/clientset/versioned"
 
 	"github.com/kubeflow/arena/pkg/apis/config"
+	"github.com/kubeflow/arena/pkg/apis/types"
 )
 
 var kubectlCmd = []string{"arena-kubectl"}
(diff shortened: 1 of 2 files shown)


pkg/util/kubectl/kubectl.go
```diff
@@ -385,7 +386,36 @@ func UpdateLWSJob(lwsJob *lwsv1.LeaderWorkerSet) error {
...
+	case string(types.VolcanoTrainingJob):
+		return "jobs.batch.volcano.sh"
+	case string(types.SparkTrainingJob):
+		return "sparkapplications.sparkoperator.k8s.io"
+	case string(types.RayJob):
+		return "rayjobs.ray.io"
+	default:
+		return ""
+	}
+}
```
ChenYi015, line 410: Since the resource name does not have suffix `s` (e.g. `pytorchjob.kubeflow.org/pytorch-mnist`), we need to remove it to avoid patching ownerReference to training job itself.

```suggestion
func getTrainingJobCRDName(trainingType string) string {
	switch trainingType {
	case string(types.TFTrainingJob):
		return "tfjob.kubeflow.org"
	case string(types.PytorchTrainingJob):
		return "pytorchjob.kubeflow.org"
	case string(types.MPITrainingJob), string(types.HorovodTrainingJob):
		return "mpijob.kubeflow.org"
	case string(types.ETTrainingJob), string(types.DeepSpeedTrainingJob):
		return "trainingjob.kai.alibabacloud.com"
	case string(types.VolcanoTrainingJob):
		return "job.batch.volcano.sh"
	case string(types.SparkTrainingJob):
		return "sparkapplication.sparkoperator.k8s.io"
	case string(types.RayJob):
		return "rayjob.ray.io"
	default:
		return ""
	}
}
```

ChenYi015 (changes requested):
@FAUST-BENCHOU Thanks for your contribution! Overall LGTM, one minor change needs to be made.

FAUST-BENCHOU:
@ChenYi015 Done.Thanks for your review!

ChenYi015 (approved):
/approve

PR #1427, opened 2026-03-03
Fix KServe minReplicas=0 Support for Scale-to-Zero

Leoyzen:
### Problem

This PR addresses two critical issues with the `minReplicas` configuration in KServe:

#### Issue 1: Helm Template Condition Fails for minReplicas=0
The Helm template used `{{- if .Values.minReplicas }}` to conditionally render the `minReplicas` field. In Helm, the value `0` is falsy, so when users set `minReplicas: 0` to enable scale-to-zero, the configuration was silently ignored and not rendered in the generated manifest.

#### Issue 2: Update Command Resets minReplicas to 1
When updating an InferenceService with any field (not `--min-replicas`), the `minReplicas` value was always reset to `1`. This occurred because:
- The default value for `--min-replicas` flag was `1` in the update command
- The update logic checked `if args.MinReplicas >= 0`, which was always true (default `1 >= 0`)
- This overwrote existing `minReplicas` settings, breaking scale-to-zero deployments

#### 1. Fixed Helm Template Condition
Changed the condition from:
```yaml
{{- if .Values.minReplicas }}
```
To:
```yaml
{{- if ge (int .Values.minReplicas) 0 }}
```

This ensures that `minReplicas=0` is correctly rendered since `0 >= 0` evaluates to `true`.

#### 2. Fixed Update Command Default Value
Changed the default value of `--min-replicas` flag in `update_serving_kserve.go` from `1` to `-1`:
```go
command.Flags().IntVar(&s.args.MinReplicas, "min-replicas", -1, "...")
```

Now:
- When `--min-replicas` is not specified, `args.MinReplicas = -1`, condition `>= 0` is `false`, no update occurs
- When `--min-replicas 0` is explicitly set, `args.MinReplicas = 0`, condition `>= 0` is `true`, updates to `0`
- When `--min-replicas <N>` (N>0) is explicitly set, updates to the specified value

### Files Changed

- `charts/kserve/templates/inferenceservice.yaml` - Fixed Helm template condition
- `charts/kserve/values.yaml` - Added default values for `minReplicas: 0` and `maxReplicas: 10`
- `pkg/argsbuilder/update_serving_kserve.go` - Changed default `--min-replicas` to `-1`

### Behavior

**Create New InferenceService:**
- Default `minReplicas` remains `1` (no breaking change)
- Users can explicitly set `--min-replicas 0` for scale-to-zero

**Update InferenceService:**
- Updating other fields no longer resets `minReplicas`
- Users can explicitly set `--min-replicas 0` to enable scale-to-zero
- Users can explicitly set `--min-replicas <N>` to change the value

### Testing Recommendations

1. Deploy a new InferenceService with `--min-replicas 0` and verify scale-to-zero works
2. Update an existing InferenceService (changing non-minReplicas fields) and verify `minReplicas` is preserved
3. Update an existing InferenceService with `--min-replicas 0` and verify it enables scale-to-zero
4. Verify existing deployments without explicit `minReplicas` continue to work as before

### Related

This fix enables true scale-to-zero functionality for KServe inference services, which is essential for cost optimization in development/test environments and low-traffic production workloads.


ChenYi015:
@Leoyzen Thanks for reporting and fixing the issue! Could you sign off your commits to pass the DCO check? See [Contributing | Kubeflow](https://www.kubeflow.org/docs/about/contributing/#sign-off-your-commits).

Leoyzen:
@ChenYi015 Done, please take a look at this pr again.

Leoyzen:
@ChenYi015 gentle ping, DCO is fixed. Could you help review or reassign?

PR #1437, opened 2026-04-28
feat(cli): add sub-command 'arena llm' as proxy to kubectl-rbg

diw-zw:
## Summary

Adds a new `llm` subcommand to the `arena` CLI that proxies all arguments to the `kubectl-rbg` binary. This allows users to access LLM deployment management capabilities through a single unified entry point (`arena llm ...`) without merging the two tools' codebases or resolving dependency conflicts.

## Motivation

`kubectl-rbg` provides LLM-related capabilities (service management, model management, benchmarking, etc.) as a separate CLI tool. Requiring users to remember and switch between two binaries degrades the user experience. This change unifies the entry point while keeping both tools independently maintainable.

## Changes

- `pkg/commands/llm/llm.go` — new file; defines `NewLLMCommand()` which delegates to `kubectl-rbg llm <args...>` via `os/exec`, with stdin/stdout/stderr passthrough and a clear error message if `kubectl-rbg` is not found in `$PATH`
- `pkg/commands/root.go` — registers the new `llm` subcommand

## Usage

```bash
# equivalent to: kubectl-rbg llm svc list
arena llm svc list

# equivalent to: kubectl-rbg llm --help
arena llm --help
```

### DIFF
--- a/pkg/commands/llm/llm.go
+++ b/pkg/commands/llm/llm.go
@@ -1,0 +1,55 @@
+// Copyright 2026 The Kubeflow Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


cheyang (approved):
/lgtm
/approve

PR #1450, opened 2026-07-04
fix(helm): process set-file options in GenerateHelmTemplate

vishalmore90:
### Context
When Arena migrated from executing the Helm CLI binary to using the Helm Go SDK (introduced in commit `19b5133e`), the `--config-file` feature was broken. Under the hood, `--config-file` generates `--set-file` parameters, but `GenerateHelmTemplate()` ignored the options slice entirely. As a result, the Helm template received `nil` instead of the file content string, causing errors such as:
`Error: template: kserve/templates/inferenceservice.yaml:423:36: executing "kserve/templates/inferenceservice.yaml" at <4>: invalid value; expected string`

This PR restores `--set-file` handling natively in the SDK template rendering path without requiring the `--helm-binary=true` workaround.

### Changes
* **[pkg/util/helm/util.go](file:///c:/Users/Vishal/Code_File/Repo%20Folder/arena/pkg/util/helm/util.go)**:
  * Parse `--set-file` arguments from the options slice.
  * Load referenced files from disk.
  * Merge file content into the nested values map using a dot-notation helper `setNestedValue`.
* **[pkg/util/helm/util_test.go](file:///c:/Users/Vishal/Code_File/Repo%20Folder/arena/pkg/util/helm/util_test.go)**:
  * Add `TestSetNestedValue` to verify type safety and correct nesting.
  * Add `TestProcessSetFileOptions` to verify command argument parsing and disk loading.

### Verification
* Unit tests in `pkg/util/helm` successfully pass:
  ```bash
  === RUN   TestSetNestedValue
  --- PASS: TestSetNestedValue (0.00s)
  === RUN   TestProcessSetFileOptions
  --- PASS: TestProcessSetFileOptions (0.01s)
  PASS
  ok  	github.com/kubeflow/arena/pkg/util/helm	4.895s


cheyang:
@gemini-code-assist /review

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid --set-file option format: %s", pair)
+		}
+		key, filePath := parts[0], parts[1]
+
+		content, err := os.ReadFile(filePath)
+		if err != nil {
+			return fmt.Errorf("failed to read file for --set-file %s: %v", filePath, err)
```
cheyang: `processSetFileOptions` passes the value portion of `--set-file key=path` directly to `os.ReadFile`. If callers ever build option strings from untrusted input, this could be used to exfiltrate or overwrite arbitrary local files. The current callers appear to pass CLI options directly, which limits the risk, but the function itself has no guardrails (no chroot, no allow-list, no symlink check).

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+			newMap := make(map[string]interface{})
+			curr[part] = newMap
+			curr = newMap
+		} else {
+			if nextMap, ok := next.(map[string]interface{}); ok {
+				curr = nextMap
+			} else if nextValues, ok := next.(chartutil.Values); ok {
+				curr = nextValues
+			} else {
+				newMap := make(map[string]interface{})
```
cheyang: When a dot-separated key traverses into an existing non-map value, `setNestedValue` replaces it with a new empty map and continues. This matches the test expectation, but it is surprising behavior and can silently discard previously-set values (including values loaded from `valuesFile`).

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+			pair = options[i+1]
+			i++
+		} else {
+			continue
+		}
+
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid --set-file option format: %s", pair)
+		}
```
cheyang: `fmt.Errorf` uses `%v` for wrapped errors (e.g. `failed to read file for --set-file %s: %v`). Using `%w` would let callers inspect the underlying error with `errors.Is` / `errors.As`.

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+		key, filePath := parts[0], parts[1]
+
+		content, err := os.ReadFile(filePath)
+		if err != nil {
+			return fmt.Errorf("failed to read file for --set-file %s: %v", filePath, err)
+		}
+
```
cheyang: `os.ReadFile` is followed by `string(content)`. For binary file inputs this will corrupt the payload. Is `--set-file` intended only for text files, and should the code reject or base64-encode binary content?

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
 	}
 	return err
 }
```
cheyang, line 249: The hand-rolled parser for `--set-file`, `--set-file=`, and `--set-file value` duplicates logic that Helm already implements. Consider delegating to Helm's `cmd/helm` or `pkg/cli/values` helpers to stay consistent with Helm's quoting/escaping rules and reduce maintenance burden.

cheyang:
Thanks for the fix. I have not approved the pending fork-gated Integration Test workflow because this PR also adds a binary artifact arena.exe that must be removed before it is safe to run CI and merge. Once the binary is removed, the --set-file implementation looks reasonable, but I left a few inline comments on input validation, silent value overwrite, and test coverage for you to consider.

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+			continue
+		}
+
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid --set-file option format: %s", pair)
+		}
+		key, filePath := parts[0], parts[1]
+
+		content, err := os.ReadFile(filePath)
```
cheyang: processSetFileOptions reads the filesystem path given in a --set-file option directly. The current caller validates config-file paths first, but this helper is low-level and could be reused by callers that do not sanitize the path. If a user-controlled option reaches this function, it becomes an arbitrary-file-read primitive. Consider requiring an absolute path, checking it is under a known config directory, or at minimum documenting that callers must validate paths.

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+		if !exists {
+			newMap := make(map[string]interface{})
+			curr[part] = newMap
+			curr = newMap
+		} else {
+			if nextMap, ok := next.(map[string]interface{}); ok {
+				curr = nextMap
+			} else if nextValues, ok := next.(chartutil.Values); ok {
+				curr = nextValues
+			} else {
```
cheyang: When a --set-file key descends through an existing scalar from values.yaml, the existing value is discarded and replaced by a new map. That can silently drop configuration that the user expected to keep. Returning an error or merging would be safer than silent replacement.

pkg/util/helm/util_test.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2026 The Kubeflow Authors.
```
cheyang, line 1: The new tests only exercise the happy path. Please add cases for: malformed --set-file option, missing file, empty key part, and nested key that overwrites an existing scalar. These tests would guard against regressions in option parsing and error handling.

pkg/util/helm/util.go
cheyang: Wrapping errors with %w lets callers inspect the underlying error with errors.Is/errors.As. The new code uses %v in several places.

pkg/util/helm/util.go
```diff
@@ -237,3 +246,70 @@ func toYaml(values interface{}, file *os.File) error {
...
+		if err := setNestedValue(values, key, string(content)); err != nil {
+			return fmt.Errorf("failed to set nested value for key %s: %v", key, err)
+		}
+	}
+	return nil
+}
+
+// setNestedValue sets a value in a nested map using dot-separated keys.
+func setNestedValue(values map[string]interface{}, key string, val interface{}) error {
+	parts := strings.Split(key, ".")
```
cheyang: Keys like configFiles[0].content will not work because setNestedValue splits only on '.'. This is fine for the current config-file keys, but it diverges from standard Helm --set-file semantics.

vishalmore90:
Hello @cheyang ,
Thank you so much for the review 
1. **Helm Parsing & Silent Overwrites**: You are right that the hand-rolled parser is limited. I will remove the custom `setNestedValue` function entirely and delegate the parsing to Helm's `strvals.ParseIntoFile`. This will natively fix the array syntax issue (e.g., `configFiles[0].content`), properly error out instead of silently overwriting existing scalars, and avoid reinventing the wheel.
2. **Security & Path Validation**: I will add explicit documentation to `processSetFileOptions` warning callers that they must validate paths before passing them, as this is a low-level helper.
3. **Error Wrapping**: I will update the `fmt.Errorf` calls to use `%w` instead of `%v` to properly wrap errors.
4. **Binary Files**: For now, this is intended primarily for text/YAML files (matching Helm's default behavior for `--set-file`), so I'll keep the string conversion but can document this limitation.
5. **Testing**: I'll add the requested unhappy path tests (missing file, malformed option, empty key part, and scalar overwrite conflicts) to ensure regressions are caught in the future.

cheyang:
@gemini-code-assist /review

cheyang:
Re-review of PR 1450: the `--set-file` logic in `pkg/util/helm/util.go` is directionally correct and the new unit tests pass, but the PR still contains two accidental artifacts that block merge: the compiled binary `arena.exe` and a scratch `main.go` at the repo root. `main.go` also breaks `go build ./...` because the root package is `arena` (see `version.go`). Please remove both files; once they are gone and CI is green, the helm fix should be ready to approve.

main.go
```diff
@@ -0,0 +1,15 @@
+package main
```
cheyang, line 1: A scratch `main.go` is added at the module root. It conflicts with the existing `version.go` (package `arena`) so `go build ./...` fails: 'found packages main (main.go) and arena (version.go)'. This file should be removed.

pkg/util/helm/util.go
```diff
@@ -237,3 +247,35 @@ func toYaml(values interface{}, file *os.File) error {
 	}
 	return err
 }
+
+// processSetFileOptions parses options for --set-file and merges the file contents into the values map.
+// Callers must ensure that the options provided are safe, particularly the file paths, to prevent arbitrary file read vulnerabilities.
+func processSetFileOptions(values map[string]interface{}, options []string) error {
```
cheyang, line 253: `processSetFileOptions` passes the option value directly to `os.ReadFile`. While this is the intended behavior for `--config-file`, the function does not restrict path traversal. Ensure callers validate `hostFile` before forming the option.

pkg/util/helm/util_test.go
```diff
@@ -0,0 +1,134 @@
...
+	t.Run("overwrite existing scalar conflict", func(t *testing.T) {
+		values := map[string]interface{}{
+			"conflictKey": "scalarValue",
+		}
+		options := []string{
+			"--set-file", "conflictKey.nested=" + filepath.ToSlash(file1Path),
+		}
+		err := processSetFileOptions(values, options)
+		if err == nil {
+			t.Fatal("expected error, got nil")
```
cheyang, line 128: The overwrite-scalar-conflict test asserts on the string 'interface conversion'. This message can change with Go or Helm versions; assert on a more stable error condition if possible.

PR #1466, opened 2026-07-14
feat: add core data model, constants, and logging

vicoooo26:
## Summary
- `pkg/constants`: framework names, CRD names, K8s labels, default images
- `pkg/log`: structured logging wrapper around klog with verbosity control
- `pkg/task`: core Task data structure, YAML loader, CLI flag overrides, `--set` expression parser (helm strvals)

Part of the Arena v2 CLI series (PR 2/7). Depends on #1465.

**Note:** This branch contains all commits from PR 1/7 (#1465). Review only the last commit (`feat: add core data model, constants, and logging`) for this PR's changes.

## Test plan
- [ ] `go test ./pkg/constants/... ./pkg/log/... ./pkg/task/... -v` passes
- [ ] `go vet ./pkg/constants/... ./pkg/log/... ./pkg/task/...` passes

PR #1465, opened 2026-07-14
chore: add arena v2 Makefile targets and gitignore patterns

vicoooo26:
## Summary
- Add Makefile with v2 build, test, vet, install, and e2e targets
- Add v2-specific patterns to .gitignore (test/e2e/crds/, .claude, /arena-v2)

Part of the Arena v2 CLI series (PR 1/7).

## Test plan
- [ ] `make arena-v2` target is defined
- [ ] `make v2-test` target is defined
- [ ] Existing code unaffected

PR #1467, opened 2026-07-14
feat: add Kubernetes client wrapper with CRD version detection

vicoooo26:
## Summary
- `pkg/client/config.go`: kubeconfig loading and REST config
- `pkg/client/kubernetes.go`: dynamic client wrapper (job CRUD, pod queries)
- `pkg/client/types.go`: JobInfo, JobStatus, and related types
- `pkg/client/crd.go`: MPIJob CRD version detection (v1alpha1 vs v1)

Part of the Arena v2 CLI series (PR 3/7). Depends on #1465, #1466.

**Note:** Review only the last commit for this PR's changes.

## Test plan
- [ ] `go test ./pkg/client/... -v` passes
- [ ] `go vet ./pkg/client/...` passes


cheyang:
@vicoooo26 — this PR is blocked from merging by a **failing DCO check**.
The DCO bot requires every commit to carry a `Signed-off-by:` trailer; without it the PR cannot land regardless of `lgtm` / `approved` labels or otherwise-green CI.

Please resolve this now. For a single-commit PR:

```
git commit --amend -s
git push --force-with-lease
```

For multiple commits on this branch:

```
git rebase --signoff origin/develop-v2
git push --force-with-lease
```

The code review side is finished and any merge labels already on the PR will take effect once DCO turns green. Reference: https://github.com/apps/dco

cheyang:
@gemini-code-assist /review

pkg/client/config.go
```diff
@@ -0,0 +1,52 @@
+package client
+
+import (
+	"github.com/kubeflow/arena/pkg/constants"
+	"k8s.io/client-go/rest"
```
cheyang, line 5: The code imports `k8s.io/client-go/rest` and other k8s packages, but the root `go.mod` has no `require` blocks and `go.sum` is missing. `go build ./pkg/client/` fails because none of the dependencies can be resolved. Please run `go mod tidy` in the PR, commit the resulting `go.mod`/`go.sum`, and verify the build succeeds before merging.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+  -X ${PACKAGE}/pkg/cli.gitCommit=${GIT_SHORT_COMMIT} \
+  -X ${PACKAGE}/pkg/cli.buildDate=${BUILD_DATE}
+
+$(LOCALBIN):
+	mkdir -p $(LOCALBIN)
+
+.PHONY: arena-v2
+arena-v2: $(LOCALBIN) ## Build arena v2 CLI for current platform.
+	@echo "Building arena v2 CLI..."
+	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) go build -ldflags '$(V2_LDFLAGS)' -o $(LOCALBIN)/arena-v2 ./cmd/arena-v2/
```
cheyang, line 49: This target builds `./cmd/arena-v2/`, but that directory does not exist in this PR. The Makefile also references `./pkg/cli/`, `./pkg/provider/`, and `./pkg/output/` in `v2-test` and `v2-vet` targets, which are likewise absent. Either add the missing packages or adjust the Makefile so the build/test targets match what is actually present.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+v2-e2e-setup: ## Download Kubeflow Training Operator and MPI Operator CRDs for v2 e2e tests.
+	@mkdir -p test/e2e/crds
+	@echo "Downloading Kubeflow Training Operator CRDs..."
+	@curl -sL https://raw.githubusercontent.com/kubeflow/trainer/v1.9.3/manifests/base/crds/kubeflow.org_pytorchjobs.yaml -o test/e2e/crds/pytorchjobs.yaml
+	@curl -sL https://raw.githubusercontent.com/kubeflow/trainer/v1.9.3/manifests/base/crds/kubeflow.org_tfjobs.yaml -o test/e2e/crds/tfjobs.yaml
+	@echo "Downloading Kubeflow MPI Operator CRDs..."
+	@curl -sL https://raw.githubusercontent.com/kubeflow/mpi-operator/v0.8.0/manifests/base/kubeflow.org_mpijobs.yaml -o test/e2e/crds/mpijobs.yaml
+	@echo "CRDs downloaded to test/e2e/crds/"
+
+.PHONY: v2-e2e-test
```
cheyang, line 97: `v2-e2e-test` is defined here, but there is no `test/e2e/` directory or any e2e tests in this PR. Since this is a feature PR introducing a new Kubernetes client wrapper and CRD version detection, we need e2e coverage that exercises the wrapper against a real cluster (or envtest) for at least create/get/list/delete and MPI version resolution.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,240 @@
...
+	"Secret":                "secrets",
+	"ServiceAccount":        "serviceaccounts",
+	"Namespace":             "namespaces",
+	"PersistentVolumeClaim": "persistentvolumeclaims",
+}
+
+// kindToGVR maps a known training job kind to its GVR.
+// Core K8s resources (ConfigMap, Pod, etc.) use the empty group.
+// MPIJob uses the resolved MPIVersion from the Client; all other kubeflow kinds use v1.
+func (c *Client) kindToGVR(kind string) (schema.GroupVersionResource, error) {
```
cheyang, line 210: `kindToGVR` hardcodes `PyTorchJob` and `TFJob` to `v1`, while only `MPIJob` goes through cluster version detection. If the cluster's storage version for PyTorchJob/TFJob is different (e.g., v2beta1), the client will submit to the wrong API version. Consider detecting the storage version for all Kubeflow CRDs, not just MPIJob.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,240 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+)
+
+// Client wraps the Kubernetes dynamic client for working with training job CRDs.
+type Client struct {
+	dynamicClient dynamic.Interface
+	MPIVersion    string // cached MPIJob API version, set by ResolveMPIVersion
```
cheyang: `MPIVersion` is a public, mutable field used as a cache, and it is read/written without synchronization. In the current CLI flow it is probably single-threaded, but if the client is ever reused across goroutines this is a data race. Consider making it private and guarding access, or documenting that the client is not safe for concurrent use.

pkg/client/crd.go
```diff
@@ -0,0 +1,102 @@
...
+		if v.Storage {
+			return v.Name
+		}
+	}
+	return ""
+}
+
+// GetCRDVersions queries the CRD object by name and returns all declared versions.
+// Returns nil, nil if the CRD is not found (k8s API returns not-found error).
+func (c *Client) GetCRDVersions(ctx context.Context, crdName string) ([]CRDVersionInfo, error) {
```
cheyang, line 48: `GetCRDVersions` returns `(nil, nil)` when the CRD is not found, which is ambiguous to callers: nil can mean either 'not found' or 'found but empty versions'. Consider returning a sentinel error or a named bool so callers don't have to inspect the nil slice to distinguish the two cases.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,240 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to delete %s %s: %w", kind, name, err)
+	}
+	return nil
+}
+
+// GetCRDObject retrieves a cluster-scoped CRD object by name.
+// Only works for apiextensions.k8s.io/v1 resources (e.g., CustomResourceDefinition).
+func (c *Client) GetCRDObject(ctx context.Context, kind, name string) (*unstructured.Unstructured, error) {
```
cheyang: `GetCRDObject` builds the resource name with `strings.ToLower(kind) + "s"`, which only works for simple English plurals. A CRD kind like `Policy` would become `policys` instead of `policies`. Since this helper is only used for `CustomResourceDefinition` today, consider either constraining the input or using the discovery client to resolve the resource name.

cheyang:
@gemini-code-assist /review

Makefile
```diff
@@ -0,0 +1,80 @@
...
+v2-e2e-test: v2-e2e-setup arena-v2 ## Run arena v2 e2e tests (requires real K8s cluster).
```
cheyang, line 98: The `v2-e2e-test` target references `test/e2e/`, but that directory does not exist in this PR. For a feature PR that adds a new Kubernetes client wrapper, CRD version detection, and the arena v2 task model, we need e2e coverage that exercises the end-to-end flow against a real or envtest-backed cluster. Unit tests alone don't validate the dynamic client integration or the MPI version resolution path. Please add e2e tests under `test/e2e/` before we can merge this.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,240 @@
...
+	"ServiceAccount":        "serviceaccounts",
+	"Namespace":             "namespaces",
+	"PersistentVolumeClaim": "persistentvolumeclaims",
+}
+
+// kindToGVR maps a known training job kind to its GVR.
+// Core K8s resources (ConfigMap, Pod, etc.) use the empty group.
+// MPIJob uses the resolved MPIVersion from the Client; all other kubeflow kinds use v1.
+func (c *Client) kindToGVR(kind string) (schema.GroupVersionResource, error) {
+	version := constants.KubeflowVersion
```
cheyang, line 211: `kindToGVR` hardcodes `KubeflowVersion` (v1) for PyTorchJob and TFJob, while only MPIJob gets storage-version detection. If a cluster installs a different storage version for PyTorchJob or TFJob CRDs, the client will send requests to the wrong version. Consider extending the CRD-version detection to all framework CRDs, or at least documenting the assumption that these two are always v1.

pkg/client/kubernetes.go
cheyang: `MPIVersion` is a public, mutable string on `Client` and acts as a cache. It is not concurrency-safe, and callers must remember to invoke `ResolveMPIVersion` before any MPIJob operation or they get an error. A cleaner design would resolve the version lazily on first MPIJob use with a `sync.Once`, or return the version alongside the client so it can't be silently stale.

pkg/client/crd.go
cheyang, line 48: `GetCRDVersions` returns `(nil, nil)` both when the CRD is not found and when `spec.versions` is missing/empty. Callers can't distinguish 'CRD not installed' from 'CRD has no versions', which makes the returned nil ambiguous. Consider returning an explicit error or a sentinel for the not-found case.

pkg/task/setter.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	result.WriteString(valueSuffix)
+	return result.String(), quotedKeys, nil
+}
+
+// restoreQuotedKeys walks a map produced by strvals.ParseInto and replaces any
+// placeholder keys with the original quoted content from the mapping. It
+// recurses into nested maps so that placeholders at any depth are restored.
+func restoreQuotedKeys(m map[string]interface{}, quotedKeys map[string]string) {
```
cheyang, line 120: `restoreQuotedKeys` deletes and inserts keys into the same map while ranging over it. Although Go permits deletion during range, adding keys while ranging is subtle and can cause a placeholder to be revisited or missed depending on map iteration. The current tests pass, but the logic is fragile; a recursive rebuild of the map would be easier to reason about.

Makefile
```diff
@@ -0,0 +1,87 @@
...
+	@mkdir -p test/e2e/crds
+	@echo "Downloading Kubeflow Training Operator CRDs..."
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/trainer/$(TRAINER_SHA)/manifests/base/crds/kubeflow.org_pytorchjobs.yaml -o test/e2e/crds/pytorchjobs.yaml
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/trainer/$(TRAINER_SHA)/manifests/base/crds/kubeflow.org_tfjobs.yaml -o test/e2e/crds/tfjobs.yaml
+	@echo "Downloading Kubeflow MPI Operator CRDs..."
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/mpi-operator/$(MPI_OPERATOR_SHA)/manifests/base/kubeflow.org_mpijobs.yaml -o test/e2e/crds/mpijobs.yaml
+	@echo "CRDs downloaded to test/e2e/crds/"
+
+.PHONY: v2-e2e-test
+v2-e2e-test: v2-e2e-setup arena-v2 ## Run arena v2 e2e tests (requires real K8s cluster).
```
cheyang, line 98: This PR introduces the Kubernetes client wrapper and MPIJob CRD version detection as a new user-facing capability, so it needs corresponding e2e coverage under `test/e2e/`. Right now `make v2-e2e-test` points at `./test/e2e/`, but no e2e tests are included in the diff, so the new `pkg/client` paths are not exercised against a real cluster. Please add e2e tests that create/list training jobs and verify the MPI version detection path.

pkg/constants/constants.go
```diff
@@ -0,0 +1,104 @@
...
+const (
+	KindPyTorchJob = "PyTorchJob"
+	KindTFJob      = "TFJob"
+	KindMPIJob     = "MPIJob"
+)
+
+// CRD group and versions.
+const (
+	KubeflowGroup     = "kubeflow.org"
+	KubeflowVersion   = "v1"
```
cheyang, line 68: `KubeflowVersion` is hardcoded to `"v1"`, and `kindToGVR` uses it for every non-MPI Kubeflow kind. That works today, but it means the client won't adapt if the cluster's PyTorchJob or TFJob CRD ever changes its storage version. Consider either querying the CRD versions for these kinds (as you do for MPIJob) or documenting why hardcoding v1 is a safe long-term assumption.

pkg/client/crd.go
```diff
@@ -0,0 +1,102 @@
...
+			Served:  served,
+			Storage: storage,
+		})
+	}
+	return result, nil
+}
+
+// ResolveMPIVersion queries the MPIJob CRD and caches the storage version in c.MPIVersion.
+// Returns immediately if MPIVersion is already set (cached).
+func (c *Client) ResolveMPIVersion(ctx context.Context) error {
```
cheyang, line 96: `ResolveMPIVersion` stores the detected MPIJob storage version in `Client.MPIVersion` and never refreshes it. For the CLI that's probably fine, but if the client is reused in a long-running process it could keep using an old version after the cluster CRD storage version changes. Consider documenting the cache lifetime or exposing a way to invalidate/recheck the version.

pkg/client/crd.go
cheyang, line 48: `GetCRDVersions` returns `nil, nil` when the CRD is not found. That's easy to confuse with an empty version list and forces every caller to distinguish by nil-check. Consider returning a sentinel error or an explicit `found` boolean so the not-found case is unambiguous at the API boundary.

go.mod
```diff
@@ -1,3 +1,49 @@
...
+	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/mailru/easyjson v0.7.7 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/spf13/pflag v1.0.10 // indirect
+	github.com/x448/float16 v0.8.4 // indirect
+	go.yaml.in/yaml/v2 v2.4.4 // indirect
```
cheyang: The module declares dependencies on `go.yaml.in/yaml/v2` and `go.yaml.in/yaml/v3` alongside the usual `gopkg.in/yaml.v3`. These look like typos and are likely to confuse dependency review or break on stricter proxies/CI. Please confirm the intended yaml module path and remove the duplicate/incorrect entries if they're not needed.

cheyang:
@gemini-code-assist /review

Makefile
```diff
@@ -0,0 +1,91 @@
...
+##@ General
+
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-30s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+##@ Arena v2
+
+# Shared package lists for v2 targets
+V2_PACKAGES := ./pkg/cli/ ./pkg/task/ ./pkg/provider/ ./pkg/client/ ./pkg/output/
```
cheyang, line 35: `V2_PACKAGES` includes `./pkg/cli/`, `./pkg/provider/`, and `./pkg/output/`, and `V2_ALL_PACKAGES` adds `./cmd/arena-v2/`. None of these directories exist at the current head, so `make v2-test` and `make v2-vet` fail with setup errors. Please either create the missing packages or remove them from the package lists until they land.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/kubeflow/arena/pkg/constants"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+)
+
+// Client wraps the Kubernetes dynamic client for working with training job CRDs.
+type Client struct {
```
cheyang, line 22: This PR introduces the Kubernetes client wrapper and CRD version detection, but there are still no e2e tests under `test/e2e/`. For a feature PR we need integration-level coverage that exercises the wrapper against a cluster (or envtest) for CRD version resolution and the Create/Get/List/Delete paths. Please add e2e tests under `test/e2e/`.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,257 @@
...
+	"PersistentVolumeClaim": "persistentvolumeclaims",
+}
+
+// kindToGVR maps a known training job kind to its GVR.
+// Core K8s resources (ConfigMap, Pod, etc.) use the empty group.
+// MPIJob uses the resolved mpiVersion from the Client; all other kubeflow kinds use v1.
+// Note: PyTorchJob and TFJob currently always use v1 in the Training Operator,
+// so hardcoding constants.KubeflowVersion ("v1") is correct for these frameworks.
+func (c *Client) kindToGVR(kind string) (schema.GroupVersionResource, error) {
+	version := constants.KubeflowVersion
```
cheyang, line 211: PyTorchJob and TFJob still hardcode `constants.KubeflowVersion` (v1). The comment explains the current Training Operator reality, but this will need to become dynamic if the operator ever serves multiple versions. Consider tracking a follow-up issue so the version detection added for MPIJob can be extended to all frameworks.

pkg/task/override.go
```diff
@@ -0,0 +1,450 @@
...
+		return err
+	}
+	if err := setStr("working-dir", &t.WorkingDir); err != nil {
+		return err
+	}
+	if err := setStr("shell", &t.Shell); err != nil {
+		return err
+	}
+
+	if v, ok := flags["run"].(string); ok && v != "" {
```
cheyang: Most flags now return a clear error on type mismatch, but `run` (line 100) and `backoff-limit` (line 294) still use bare type assertions. Making them consistent with `setStr`/`setInt` would give users clearer CLI error messages.

PR #1468, opened 2026-07-14
feat: add job providers for MPI, PyTorch, and TensorFlow

vicoooo26:
## Summary
- `pkg/provider/interface.go`: Provider interface definition
- `pkg/provider/mpi.go`: MPIJob provider (supports v1alpha1 and v1 CRD versions)
- `pkg/provider/pytorch.go`: PyTorchJob provider
- `pkg/provider/tensorflow.go`: TFJob provider
- Volume building and container mount resolution utilities

Part of the Arena v2 CLI series (PR 4/7). Depends on #1465, #1466, #1467.

**Note:** Review only the last commit for this PR's changes.

## Test plan
- [ ] `go test ./pkg/provider/... -v` passes
- [ ] `go vet ./pkg/provider/...` passes


cheyang:
@vicoooo26 — this PR is blocked from merging by a **failing DCO check**.
The DCO bot requires every commit to carry a `Signed-off-by:` trailer; without it the PR cannot land regardless of `lgtm` / `approved` labels or otherwise-green CI.

Please resolve this now. For a single-commit PR:

```
git commit --amend -s
git push --force-with-lease
```

For multiple commits on this branch:

```
git rebase --signoff origin/develop-v2
git push --force-with-lease
```

The code review side is finished and any merge labels already on the PR will take effect once DCO turns green. Reference: https://github.com/apps/dco

cheyang:
@gemini-code-assist /review

pkg/provider/interface.go
```diff
@@ -0,0 +1,1047 @@
...
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/kubeflow/arena/pkg/constants"
+	"github.com/kubeflow/arena/pkg/task"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
```
cheyang, line 14: This file imports `k8s.io/apimachinery/pkg/apis/meta/v1/unstructured`, but `go.mod` only declares the module and `go 1.25.1` with no `require` blocks, and there is no `go.sum`. The same applies to `k8s.io/client-go`, `gopkg.in/yaml.v3`, `helm.sh/helm/v3`, `github.com/stretchr/testify`, and `k8s.io/klog/v2`. Running `go test ./pkg/provider/...` fails immediately with `no required module provides package`. Please run `go mod tidy` and commit the resulting `go.mod` + `go.sum` so the project builds and CI can run.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+  -X ${PACKAGE}/pkg/cli.gitCommit=${GIT_SHORT_COMMIT} \
+  -X ${PACKAGE}/pkg/cli.buildDate=${BUILD_DATE}
+
+$(LOCALBIN):
+	mkdir -p $(LOCALBIN)
+
+.PHONY: arena-v2
+arena-v2: $(LOCALBIN) ## Build arena v2 CLI for current platform.
+	@echo "Building arena v2 CLI..."
+	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) go build -ldflags '$(V2_LDFLAGS)' -o $(LOCALBIN)/arena-v2 ./cmd/arena-v2/
```
cheyang, line 49: The `arena-v2` target builds `./cmd/arena-v2/`, but no `cmd/arena-v2` directory is present in this PR. `v2-test` and `v2-vet` also reference `./pkg/cli/` and `./pkg/output/`, which are missing. Please align the Makefile with the actual package layout, or add the missing packages, so the build and test targets do not fail for the wrong reasons.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+
+.PHONY: v2-test
+v2-test: ## Run arena v2 unit tests.
+	@echo "Running arena v2 unit tests..."
+	go test ./pkg/cli/ ./pkg/task/ ./pkg/provider/ ./pkg/client/ ./pkg/output/ -v
```
cheyang: The `v2-test` target lists `./pkg/cli/` and `./pkg/output/`, neither of which exists in this PR. `v2-vet` also references `./pkg/cli/` and `./cmd/arena-v2/`. Please align the Makefile with the actual package layout, otherwise `make v2-test` and `make v2-vet` will fail for the wrong reasons.

Makefile
```diff
@@ -0,0 +1,80 @@
...
+v2-e2e-setup: ## Download Kubeflow Training Operator and MPI Operator CRDs for v2 e2e tests.
+	@mkdir -p test/e2e/crds
+	@echo "Downloading Kubeflow Training Operator CRDs..."
+	@curl -sL https://raw.githubusercontent.com/kubeflow/trainer/v1.9.3/manifests/base/crds/kubeflow.org_pytorchjobs.yaml -o test/e2e/crds/pytorchjobs.yaml
+	@curl -sL https://raw.githubusercontent.com/kubeflow/trainer/v1.9.3/manifests/base/crds/kubeflow.org_tfjobs.yaml -o test/e2e/crds/tfjobs.yaml
+	@echo "Downloading Kubeflow MPI Operator CRDs..."
+	@curl -sL https://raw.githubusercontent.com/kubeflow/mpi-operator/v0.8.0/manifests/base/kubeflow.org_mpijobs.yaml -o test/e2e/crds/mpijobs.yaml
+	@echo "CRDs downloaded to test/e2e/crds/"
+
+.PHONY: v2-e2e-test
```
cheyang, line 97: This is a `feat:` PR that introduces MPI, PyTorch, and TensorFlow job providers. The project requires corresponding e2e tests under `test/e2e/` that exercise CRD creation through the Kubernetes API. The `v2-e2e-test` Makefile target references `test/e2e/`, but that directory does not exist. Please add e2e coverage for at least one happy path per framework plus a version-resolution case for MPIJob.

pkg/task/loader.go
```diff
@@ -0,0 +1,29 @@
+package task
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
```
cheyang, line 7: `go.mod` declares `go 1.25.1`, which is not a released Go version (the latest stable release is in the 1.24 series). Combined with the missing `require` blocks and absent `go.sum`, this prevents any build or test run. Please set the directive to a released version such as `1.22` or `1.23` and run `go mod tidy`.

cheyang:
@gemini-code-assist /review

Makefile
cheyang, line 97: This PR introduces three new training-job providers (MPI, PyTorch, TensorFlow) but does not include any e2e tests under `test/e2e/`. For feature PRs that add user-facing controller/CRD behavior, unit tests alone are not sufficient: we need tests that exercise the full create-CRD-and-submit path against a real or envtest-backed Kubernetes API. Please add e2e coverage that builds each provider's CRD and validates it against the corresponding Training Operator CRD schema.

Makefile
cheyang: `make v2-test`, `make v2-vet`, and `make arena-v2` reference `./pkg/cli/`, `./pkg/output/`, and `./cmd/arena-v2/`, but none of these packages are included in this PR. Checking out this branch and running any of those targets fails immediately with `no Go files in ...`. Either include those packages in this PR or narrow the Makefile targets to the packages that actually exist here (`pkg/provider`, `pkg/task`, `pkg/client`, `pkg/constants`, `pkg/log`).

pkg/task/types.go
```diff
@@ -0,0 +1,612 @@
...
+	}
+
+	return nil
+}
+
+// EffectiveShell returns the shell to use, defaulting to /bin/sh for
+// empty or invalid values.
+func (t *Task) EffectiveShell() string {
+	s := strings.TrimSpace(t.Shell)
+	if s == "" || !strings.HasPrefix(s, "/") {
```
cheyang: `EffectiveShell` treats any shell that does not start with `/` as invalid and silently falls back to `/bin/sh`. That means a Task with `shell: bash` or `shell: sh` will be rewritten to `/bin/sh`, which is surprising and breaks user intent. Accept bare executable names or document that only absolute paths are allowed; otherwise fix the fallback logic.

pkg/provider/interface.go
```diff
@@ -0,0 +1,1047 @@
...
+	}
+	sp := map[string]interface{}{}
+	if s.Queue != "" {
+		sp["queue"] = s.Queue
+	}
+	if s.PriorityClassName != "" {
+		sp["priorityClass"] = s.PriorityClassName
+	}
+	if s.Gang.Enabled {
+		sp["minAvailable"] = int64(0)
```
cheyang: When gang scheduling is enabled, `buildSchedulingPolicy` emits `minAvailable: 0`. A value of zero defeats gang scheduling because the scheduler will not wait for any pods before launching the job. Compute the actual total replica count (worker + master/chief/launcher/ps where applicable) and set `minAvailable` to that value.

pkg/provider/interface.go
```diff
@@ -0,0 +1,1047 @@
...
+	// Extract last path component
+	parts := strings.Split(name, "/")
+	if len(parts) > 0 {
+		return parts[len(parts)-1]
+	}
+	return "repo"
+}
+
+// buildMetadata creates a metadata map with name, labels, and annotations.
+func buildMetadata(t *task.Task) map[string]interface{} {
```
cheyang, line 913: `buildMetadata` builds `name`, `labels`, and `annotations` but never sets `namespace`. The resulting `unstructured.Unstructured` object has an empty namespace, so `Client.Create` will submit it to whatever namespace was last used by the dynamic client or fail validation. Set `metadata.namespace` from `t.Namespace` in each provider's `BuildCRD`.

Makefile
```diff
@@ -0,0 +1,87 @@
...
+  -X ${PACKAGE}/pkg/cli.gitCommit=${GIT_SHORT_COMMIT} \
+  -X ${PACKAGE}/pkg/cli.buildDate=${BUILD_DATE}
+
+$(LOCALBIN):
+	mkdir -p $(LOCALBIN)
+
+.PHONY: arena-v2
+arena-v2: $(LOCALBIN) ## Build arena v2 CLI for current platform.
+	@echo "Building arena v2 CLI..."
+	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) go build -ldflags '$(V2_LDFLAGS)' -o $(LOCALBIN)/arena-v2 ./cmd/arena-v2/
```
cheyang, line 49: The Makefile points to `./cmd/arena-v2/`, `./pkg/cli/`, and `./pkg/output/`, but none of these directories are present in the PR. Running `make arena-v2`, `make v2-test`, `make v2-vet`, or `make v2-install` will fail immediately with package-not-found errors.

Makefile
```diff
@@ -0,0 +1,87 @@
...
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/trainer/$(TRAINER_SHA)/manifests/base/crds/kubeflow.org_pytorchjobs.yaml -o test/e2e/crds/pytorchjobs.yaml
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/trainer/$(TRAINER_SHA)/manifests/base/crds/kubeflow.org_tfjobs.yaml -o test/e2e/crds/tfjobs.yaml
+	@echo "Downloading Kubeflow MPI Operator CRDs..."
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/mpi-operator/$(MPI_OPERATOR_SHA)/manifests/base/kubeflow.org_mpijobs.yaml -o test/e2e/crds/mpijobs.yaml
+	@echo "CRDs downloaded to test/e2e/crds/"
+
+.PHONY: v2-e2e-test
+v2-e2e-test: v2-e2e-setup arena-v2 ## Run arena v2 e2e tests (requires real K8s cluster).
+	@echo "Running arena v2 e2e tests..."
+	go test -tags v2e2e ./test/e2e/ -v -ginkgo.v -timeout 30m
```
cheyang, line 100: This PR introduces new MPI/PyTorch/TensorFlow job providers, but there is no `test/e2e/` directory and no v2 e2e tests. The `v2-e2e-test` target references `test/e2e/`, which does not exist.

pkg/provider/interface.go
cheyang, line 913: `buildMetadata` sets `name`, `labels`, and `annotations` but never applies `Task.Namespace` to the generated CRD. The client's `Create` then calls `crd.GetNamespace()` and creates the object in an empty namespace, which will be rejected by the API server or default unexpectedly.

pkg/provider/interface.go
cheyang: When gang scheduling is enabled, `buildSchedulingPolicy` hard-codes `minAvailable` to 0. A zero value means no gang-allocation guarantee, defeating the purpose of enabling gang scheduling.

pkg/task/types.go
cheyang: `EffectiveShell` treats any shell not starting with `/` as invalid and silently falls back to `/bin/sh`. A user who writes `shell: bash` will have their command run under `/bin/sh` without warning.

pkg/client/crd.go
```diff
@@ -0,0 +1,102 @@
...
+
+	versions, err := c.GetCRDVersions(ctx, "mpijobs.kubeflow.org")
+	if err != nil {
+		return fmt.Errorf("failed to get MPIJob CRD versions: %w", err)
+	}
+	if versions == nil {
+		return fmt.Errorf("MPIJob CRD not found in cluster")
+	}
+
+	storageVersion := FindStorageVersion(versions)
```
cheyang, line 109: `ResolveMPIVersion` picks whatever version is marked as the CRD storage version and assigns it directly to `c.MPIVersion`. It does not check whether that version is in `mpiSupportedVersions`.

cheyang:
@gemini-code-assist /review

cheyang:
Thanks for fixing the DCO issue. However, several blocking issues from the previous round are still unresolved: the Makefile still references missing packages (pkg/cli, pkg/output, cmd/arena-v2) and the v2 e2e target points to a non-existent test/e2e/ directory. I also found a new correctness issue where MPI gang scheduling undercounts the implicit launcher replica. I've left inline comments on each. Please address these and re-push.

Makefile
```diff
@@ -0,0 +1,91 @@
...
+##@ General
+
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-30s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+##@ Arena v2
+
+# Shared package lists for v2 targets
+V2_PACKAGES := ./pkg/cli/ ./pkg/task/ ./pkg/provider/ ./pkg/client/ ./pkg/output/
```
cheyang, line 35: The Makefile's v2 targets reference ./pkg/cli/, ./pkg/output/, and ./cmd/arena-v2/, but none of these packages are present in the PR (there is no cmd/ directory at all). Because of this, make arena-v2, make v2-vet, make v2-install, and the integration-test target will fail with "no such file or directory". The prior review round already flagged this; it appears to be unresolved.

Makefile
```diff
@@ -0,0 +1,91 @@
...
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/trainer/$(TRAINER_SHA)/manifests/base/crds/kubeflow.org_pytorchjobs.yaml -o test/e2e/crds/pytorchjobs.yaml
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/trainer/$(TRAINER_SHA)/manifests/base/crds/kubeflow.org_tfjobs.yaml -o test/e2e/crds/tfjobs.yaml
+	@echo "Downloading Kubeflow MPI Operator CRDs..."
+	@curl -fsSL https://raw.githubusercontent.com/kubeflow/mpi-operator/$(MPI_OPERATOR_SHA)/manifests/base/kubeflow.org_mpijobs.yaml -o test/e2e/crds/mpijobs.yaml
+	@echo "CRDs downloaded to test/e2e/crds/"
+
+.PHONY: v2-e2e-test
+v2-e2e-test: v2-e2e-setup arena-v2 ## Run arena v2 e2e tests (requires real K8s cluster).
+	@echo "Running arena v2 e2e tests..."
+	go test -tags v2e2e ./test/e2e/ -v -ginkgo.v -timeout 30m
```
cheyang, line 100: The v2-e2e-test target runs go test -tags v2e2e ./test/e2e/, but the test/e2e/ directory does not exist and no e2e tests are included. For a feature PR that introduces three new job providers, unit tests alone are not sufficient; we need e2e coverage that exercises at least one provider through the Kubernetes API surface (envtest or Kind).

pkg/provider/interface.go
```diff
@@ -0,0 +1,1096 @@
...
+	if t.Chief != nil {
+		total++
+	}
+	if t.PS != nil {
+		total += int64(t.PS.Replicas)
+	}
+	if t.Evaluator != nil {
+		total++
+	}
+	if t.Launcher != nil {
```
cheyang, line 389: totalReplicas only counts the launcher role when t.Launcher is explicitly configured, but MPIProvider.BuildCRD always injects a launcher replica through buildLauncherSpec even when t.Launcher is nil. As a result, an MPI job with gang scheduling enabled gets runPolicy.schedulingPolicy.minAvailable equal to the worker count instead of workers + launcher. The scheduler can then place the launcher before all workers are available, defeating the all-or-nothing gang guarantee.

pkg/provider/mpi_rbac.go
```diff
@@ -0,0 +1,110 @@
...
+			"resources": []interface{}{"pods"},
+		},
+	}
+
+	if workerReplicas > 0 {
+		// Pod names follow the training-operator convention: <jobName>-<replicaType>-<index>.
+		// The MPI operator creates worker pods named <jobName>-worker-0, <jobName>-worker-1, etc.
+		// This must stay in sync with the operator's naming scheme; if the operator changes
+		// its naming convention, this list will need to be updated accordingly.
+		podNames := make([]interface{}, workerReplicas)
```
cheyang, line 50: The Role's pods/exec resourceNames are generated as <jobName>-worker-<index>, assuming the training-operator names worker pods exactly that way. If the operator's naming convention changes (or differs between MPIJob v1 and v2beta1), the launcher will lose exec permission on workers. Keep this assumption in sync with the operator version detection logic, or consider a broader but still namespaced permission model if the convention is not stable.

pkg/provider/interface_test.go
```diff
@@ -0,0 +1,1229 @@
...
+	}
+	if result["queue"] != "high-priority" {
+		t.Errorf("expected queue=high-priority, got %v", result["queue"])
+	}
+	if result["priorityClass"] != "premium" {
+		t.Errorf("expected priorityClass=premium, got %v", result["priorityClass"])
+	}
+}
+
+func TestBuildSchedulingPolicyGang(t *testing.T) {
```
cheyang, line 70: TestBuildSchedulingPolicyGang currently covers a generic master+worker case. Please add an MPI-specific case where the launcher is implicitly created (t.Launcher == nil) and assert that minAvailable equals workers + 1.

PR #1473, opened 2026-07-17
feat(arena-artifacts): mount init container ConfigMap into pytorch-operator

vicoooo26:
## Purpose of this PR

Bring the `pytorch-operator` subchart in `arena-artifacts` in line with the newer `training-operator` chart: use a modern Alpine image for the init container and expose the init container template as a file on disk instead of relying on the operator's built-in Go template. This lets cluster admins patch the template (e.g. DNS wait-loop tuning) without rebuilding the operator image.

**Proposed changes:**

- Bump `arena-artifacts` `pytorch.initContainerImage` from `alpine:3.10` to `alpine:3.22.2`.
- Add a new ConfigMap `pytorch-init-container-template` in the pytorch-operator subchart, containing `initContainer.yaml` with the DNS wait-loop template (matching the reference `training-operator` chart).
- Mount the ConfigMap read-only at `/etc/config` on the `pytorch-operator` container so the operator reads `/etc/config/initContainer.yaml` at runtime. Template placeholders `{{.InitContainerImage}}`, `{{.MaxTries}}`, and `{{.MasterAddr}}` are kept as literals in the ConfigMap and substituted by the operator's own Go template per PyTorchJob.

## Change Category

- [ ] Documentation update

### Rationale

The upstream `training-operator` chart already ships this pattern (see `aliyun-incubator/training-operator` reference). Mirroring it in the `pytorch-operator` subchart keeps the two operator deployments consistent, upgrades the stale `alpine:3.10` image to a maintained `alpine:3.22.2`, and moves the init container template out of the binary into a ConfigMap that operators can edit in place.

## Test Plan

- `helm template arena-artifacts ./arena-artifacts --namespace arena-system` renders cleanly and shows:
  - a `ConfigMap` named `pytorch-init-container-template` with the expected `initContainer.yaml` body and literal `{{...}}` placeholders;
  - the `pytorch-operator` `Deployment` with `volumeMounts: /etc/config` and the matching `volumes` entry;
  - the operator container arg `--init-container-image=alpine:3.22.2`.
- Deploy to a test cluster and confirm the pytorch-operator pod starts with `/etc/config/initContainer.yaml` mounted and successfully spawns PyTorchJob init containers using the file-based template.

### DIFF
--- a/arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
+++ b/arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.initContainerTemplateEnabled -}}
+{{- $defaultResources := dict "limits" (dict "cpu" "100m" "memory" "20Mi") "requests" (dict "cpu" "50m" "memory" "10Mi") -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: pytorch-init-container-template
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "arena.labels" . | nindent 4 }}
+data:
+  initContainer.yaml: |
+    - name: init-pytorch
+      image: {{"{{"}}.InitContainerImage{{"}}"}}
+      imagePullPolicy: {{ .Values.initContainerImagePullPolicy | default "IfNotPresent" }}
+      resources:
+        {{- toYaml (.Values.initContainerResources | default $defaultResources) | nindent 8 }}
+      command: ['sh', '-c', 'err=1;for i in $(seq {{ .Values.initContainerMaxTries | default 150 }}); do if getent hosts {{"{{"}}.MasterAddr{{"}}"}}; then err=0 && break; fi;echo waiting for master; sleep {{ .Values.initContainerSleepSeconds | default 2 }}; done; exit $err']
+{{- end -}}
--- a/arena-artifacts/charts/pytorch-operator/templates/operator-dp.yaml
+++ b/arena-artifacts/charts/pytorch-operator/templates/operator-dp.yaml
@@ -45,4 +45,16 @@ spec:
           name: pytorch-operator
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          {{- if .Values.initContainerTemplateEnabled }}
+          volumeMounts:
+            - mountPath: /etc/config
+              name: init-container-template
+              readOnly: true
+          {{- end }}
       serviceAccountName: pytorch-operator
+      {{- if .Values.initContainerTemplateEnabled }}
+      volumes:
+        - name: init-container-template
+          configMap:
+            name: pytorch-init-container-template
+      {{- end }}
--- a/arena-artifacts/tests/pytorch-operator/configmap_test.yaml
+++ b/arena-artifacts/tests/pytorch-operator/configmap_test.yaml
@@ -1,0 +1,162 @@
+#
+# Copyright 2025 The Kubeflow authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+suite: Test pytorch operator configmap
+
+templates:
+- charts/pytorch/templates/operator-configmap.yaml
+
+release:
+  name: arena-artifacts
+  namespace: arena-system
+
+set:
+  pytorch:
+    enabled: true
+
+tests:
+- it: Should not render ConfigMap when initContainerTemplateEnabled is false
+  set:
+    pytorch:
+      initContainerTemplateEnabled: false
+  asserts:
+  - hasDocuments:
+      count: 0
+
+- it: Should render ConfigMap with correct apiVersion and kind
+  asserts:
+  - equal:
+      path: apiVersion
+      value: v1
+  - equal:
+      path: kind
+      value: ConfigMap
+
+- it: Should render ConfigMap with correct name and namespace
+  asserts:
+  - equal:
+      path: metadata.name
+      value: pytorch-init-container-template
+  - equal:
+      path: metadata.namespace
+      value: arena-system
+
+- it: Should render ConfigMap with arena labels
+  asserts:
+  - equal:
+      path: metadata.labels["helm.sh/chart"]
+      value: arena-artifacts
+  - equal:
+      path: metadata.labels["app.kubeflow.org/managed-by"]
+      value: arena
+
+- it: Should preserve Go template literals for operator-side substitution
+  asserts:
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "\\{\\{\\.InitContainerImage\\}\\}"
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "\\{\\{\\.MasterAddr\\}\\}"
+
+- it: Should render default values from values.yaml
+  asserts:
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "imagePullPolicy: IfNotPresent"
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "seq 150"
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "sleep 2"
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "cpu: 100m"
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
+      pattern: "memory: 20Mi"
+  - matchRegex:
+      path: 'data["initContainer.yaml"]'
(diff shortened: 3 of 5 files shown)


arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
```diff
@@ -0,0 +1,20 @@
...
+      image: {{"{{"}}.InitContainerImage{{"}}"}}
+      imagePullPolicy: IfNotPresent
+      resources:
+        limits:
+          cpu: 100m
+          memory: 20Mi
+        requests:
+          cpu: 50m
+          memory: 10Mi
+      command: ['sh', '-c', 'err=1;for i in $(seq 100); do if getent hosts {{"{{"}}.MasterAddr{{"}}"}}; then err=0 && break; fi;echo waiting for master; sleep 2; done; exit $err']
```
cheyang: This retry loop is bounded to `seq 100`, so the init container gives up after about 200s (100 iterations x `sleep 2`). That feels a bit tight when the master pod is slow to schedule or still pulling a large image, and the job's init would then fail instead of waiting a little longer. Could we default this to 300s by bumping it to `seq 150`? The operator only substitutes `.MasterAddr` and `.InitContainerImage`, so the count can't be templated per-job, but keeping it in the ConfigMap at least lets admins tune it later.

cheyang:
@gemini-code-assist /review

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
vicoooo26: Bumped to 150. It's worth making those values configurable in our trainer (training-operator) chart.

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
cheyang, line 3: This PR introduces a new Helm template (`operator-configmap.yaml`) and adds a new `volumeMounts`/`volumes` pair to `operator-dp.yaml`, but no helm-unittest coverage is added for either change. The project runs `make helm-unittest` in CI, and the repo convention requires e2e/unit tests for new features. Please add tests under `arena-artifacts/tests/pytorch-operator/` that assert the ConfigMap is rendered with the expected `initContainer.yaml` content and that the Deployment mounts it at `/etc/config` read-only.

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: ConfigMap
+metadata:
+  name: pytorch-init-container-template
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "arena.labels" . | nindent 4 }}
+data:
+  initContainer.yaml: |
+    - name: init-pytorch
```
cheyang, line 12: The init container template hard-codes `resources` (cpu/memory limits and requests) and `imagePullPolicy: IfNotPresent`. These values may not suit all clusters and are not configurable via `values.yaml`. Consider exposing them through Helm values so operators can tune them without editing the chart.

arena-artifacts/charts/pytorch-operator/templates/operator-dp.yaml
```diff
@@ -45,4 +45,12 @@ spec:
           name: pytorch-operator
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
```
cheyang, line 49: Please add a helm-unittest case that renders `operator-dp.yaml` and asserts the `init-container-template` volume is mounted at `/etc/config` with `readOnly: true` and that the corresponding `volumes` entry references `pytorch-init-container-template`.

cheyang:
@gemini-code-assist /review

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
```diff
@@ -0,0 +1,16 @@
+{{- $defaultResources := dict "limits" (dict "cpu" "100m" "memory" "20Mi") "requests" (dict "cpu" "50m" "memory" "10Mi") -}}
```
cheyang, line 2: This PR introduces a new user-facing capability (mounting an init-container template ConfigMap into pytorch-operator so PyTorchJobs use it), but it does not add any e2e tests under `test/e2e/`. Per the project convention for feature PRs, e2e tests are required to exercise the end-to-end behavior through the Kubernetes API surface.

arena-artifacts/tests/pytorch-operator/configmap_test.yaml
```diff
@@ -0,0 +1,117 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+suite: Test pytorch operator configmap
+
+templates:
+- charts/pytorch/templates/operator-configmap.yaml
```
cheyang, line 20: The `templates` list points to `charts/pytorch/templates/operator-configmap.yaml`, but the actual subchart directory is `charts/pytorch-operator/`. This mismatch means the test suite will fail to load the intended template.

arena-artifacts/charts/pytorch-operator/templates/operator-dp.yaml
```diff
@@ -45,4 +45,12 @@ spec:
           name: pytorch-operator
```
cheyang, line 45: The Deployment now mounts the ConfigMap at `/etc/config`, but the chart still passes `--init-container-image={{ .Values.initContainerImage }}` to the operator. The feature depends on the `acs/pytorch-operator:2845db5-aliyun` image reading `/etc/config/initContainer.yaml`.

arena-artifacts/values.yaml
```diff
@@ -117,7 +126,7 @@ et:
       cpu: 100m
```
cheyang, line 128: The PR title and description focus on pytorch-operator, but the diff also bumps `et.initContainerImage` from `alpine:3.10` to `alpine:3.22.2`. The change is reasonable for security, but it is not covered by the new tests and is unrelated to the feature.

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: ConfigMap
+metadata:
+  name: pytorch-init-container-template
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "arena.labels" . | nindent 4 }}
+data:
+  initContainer.yaml: |
+    - name: init-pytorch
+      image: {{"{{"}}.InitContainerImage{{"}}"}}
```
cheyang, line 13: The ConfigMap preserves the literal strings `{{.InitContainerImage}}` and `{{.MasterAddr}}` for the operator to substitute later. I want to confirm the operator uses Go templating (or equivalent) on the file contents at job creation time.

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
vicoooo26: `arena-artifacts` is an internal Helm Chart installed by a bash script. As there is no Go code involved, we only need Helm Chart tests.

arena-artifacts/tests/pytorch-operator/configmap_test.yaml
vicoooo26: There are aliases for each subchart of `arena-artifacts` in `Chart.yaml` that will still remain effective during Helm unittest.

arena-artifacts/charts/pytorch-operator/templates/operator-dp.yaml
vicoooo26: The `init-container` feature is not introduced in `acs/pytorch-operator:2845db5-aliyun`. It is still supported in older images.

arena-artifacts/values.yaml
vicoooo26: The ET `init-container` just [copies and changes permissions of some host files](https://github.com/AliyunContainerService/et-operator/blob/96806fbbd1f8f4c067661a2d3ec8780b874c62a5/pkg/controllers/resource.go#L173). Since almost all Alpine distros support `cp` and `chmod`, this works fine.

arena-artifacts/charts/pytorch-operator/templates/operator-configmap.yaml
vicoooo26: `{{.InitContainerImage}}` and `{{.MasterAddr}}` are used in operator [templating logic](https://github.com/AliyunContainerService/pytorch-operator/blob/2845db57cb7ca2a39038796830b436cf645be3ef/pkg/controller.v1/pytorch/pod.go#L189).

cheyang (approved):
/lgtm
/approve

arena-artifacts/values.yaml
```diff
@@ -100,7 +100,18 @@ pytorch:
...
       memory: 256Mi
   nodeSelector: {}
-  initContainerImage: alpine:3.10
+  initContainerImage: alpine:3.22.2
+  initContainerTemplateEnabled: true
```
cheyang, line 104: This feature PR introduces a new init-container ConfigMap and mounts it into the PyTorch operator, but there is still no e2e test under `test/e2e/` exercising that behavior. For new user-facing features we expect an integration-level test (for example, validating that the rendered operator Deployment includes the `/etc/config` volume and that the ConfigMap is created). Could you add coverage here so regressions are caught beyond helm unit tests?

PR #1475, opened 2026-07-19
fix: resolve real username for client-certificate kubeconfigs

darrencrosz:
Arena previously hard-coded arena.kubeflow.org/username to kubecfg:certauth:admin whenever kubeconfig used client certs. That breaks multi-user ACK/RAM X509 setups where the certificate CN is the real principal.

Prefer SelfSubjectReview (same as kubectl auth whoami), then fall back to parsing the client certificate CommonName, and only then keep the legacy admin placeholder.

## Purpose of this PR

**Proposed changes:**

- <Change 1>
- <Change 2>
- <Change 3>

## Change Category

- [ ] Documentation update

### DIFF
--- a/pkg/apis/config/user.go
+++ b/pkg/apis/config/user.go
@@ -16,10 +16,12 @@ package config
 
 import (
 	"context"
-	"errors"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"os"
 
+	log "github.com/sirupsen/logrus"
 	authenticationapi "k8s.io/api/authentication/v1"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


cheyang:
@gemini-code-assist /review

pkg/apis/config/user.go
```diff
@@ -38,6 +40,21 @@ func getUserName(namespace string, clientConfig clientcmd.ClientConfig, restConf
 		return &userName, nil
 	}
 	if tc.HasCertAuth() {
+		// Prefer asking the API server who we are (same as `kubectl auth whoami`).
+		// Historically this branch hard-coded "kubecfg:certauth:admin", which is wrong for
+		// per-user client certificates (e.g. ACK RAM X509 kubeconfigs whose CN is the RAM UID).
+		if userName, err := getUserNameBySelfSubjectReview(clientSet); err == nil {
```
cheyang, line 46: The new primary resolution path uses `authentication.k8s.io/v1.SelfSubjectReview`, which only graduated to v1 in Kubernetes 1.28 and was available as v1beta1 in 1.26. On older clusters this call will fail and the code correctly falls back to certificate CN parsing, then to the legacy admin placeholder. This is handled gracefully, but worth noting in release notes for users on older Kubernetes versions.

pkg/apis/config/user_test.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+	}
+	der, err := x509.CreateCertificate(rand.Reader, tmpl, tmpl, &key.PublicKey, key)
+	if err != nil {
+		t.Fatalf("create certificate: %v", err)
+	}
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: der})
+}
+
+func TestGetUserNameFromClientCertificateCertData(t *testing.T) {
```
cheyang, line 57: The new tests cover the certificate-CN fallback path well, but the primary `getUserNameBySelfSubjectReview` path is not exercised. Adding a test with a fake `kubernetes.Interface` that returns a mocked `SelfSubjectReview` would guard the main resolution path against regressions.

pkg/apis/config/user.go
```diff
@@ -82,12 +99,57 @@ func getUserNameByToken(kubeclient kubernetes.Interface, token string) (*string,
...
+		var err error
+		certPEM, err = os.ReadFile(restConfig.CertFile)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("no client certificate data or file configured")
+	}
+
+	block, _ := pem.Decode(certPEM)
```
cheyang, line 138: `pem.Decode` only parses the first PEM block in `CertData`. If a user supplies a full certificate chain, the first block might not be the leaf with the intended CommonName. Iterating through PEM blocks until a certificate with a non-empty CommonName is found, or explicitly documenting that only the first certificate is considered, would make this more robust.

cheyang (approved):
All CI checks passed and maintainer has approved. Auto-approving to satisfy branch protection.

PR #1483, opened 2026-07-20
feat: add example configurations and end-to-end tests

vicoooo26:
## Summary
- 11 example YAML files: MPI (simple, with-launcher), PyTorch (simple, standalone, tensorboard-mounts, with-master), TensorFlow (simple, with-roles), test fixtures (configmap, features, secret)
- E2E tests: suite setup, MPI/PyTorch/TFJob submission, dry-run, format, check, CLI overrides, YAML submit/delete
- Integration tests: flow, loader, multijob, override, output, stop/resume, validation, error display

Part of the Arena v2 CLI series (PR 7/7). Depends on #1477, #1478, #1479, #1480, #1481, #1482.

**Note:** Review only the last commit for this PR's changes.

## Previous PR
This is a resubmission of the closed PR #1471. All review comments from #1471 have been addressed and fixed in this branch.

### Review comments addressed:
- Qualified all default images with `docker.io/` registry prefix (rsync, HDFS, TensorBoard) for consistency with `DefaultGitSyncImage` which uses `registry.k8s.io/`
- Qualified all image references in example YAMLs and test fixtures with explicit registry prefixes
- Fixed CLAUDE.md build instructions to use `arena-v2` consistently
- All other review comments on examples and tests have been resolved

## Test plan
- [ ] `go test ./test/... -v` passes (integration tests)
- [ ] `go vet ./test/...` passes
- [ ] E2E tests compile: `go test -tags v2e2e ./test/e2e/... -v -run None`
- [ ] Example YAMLs are valid: `go run ./cmd/arena-v2/ submit --dry-run examples/v2/pytorch-simple.yaml`

PR #1482, opened 2026-07-20
feat: add CLI extended commands for lifecycle and observation

vicoooo26:
## Summary
- `pkg/cli/logs.go`: `arena job logs` — view pod logs
- `pkg/cli/delete.go`: `arena job delete` — delete jobs
- `pkg/cli/suspend.go`, `resume.go`: `arena job suspend/resume` — lifecycle management
- `pkg/cli/top.go`: `arena top job` — resource usage monitoring
- `pkg/cli/check.go`: `arena check` — environment and CRD verification
- `pkg/cli/integration_test.go`: integration tests (build-tagged)

Part of the Arena v2 CLI series (PR 6/7). Depends on #1477, #1478, #1479, #1480, #1481.

**Note:** Review only the last commit for this PR's changes.

## Previous PR
This is a resubmission of the closed PR #1470. All review comments from #1470 have been addressed and fixed in this branch.

### Review comments addressed:
- All review comments from the closed PR #1470 have been resolved, including feedback on label selector safety, test coverage, extended CLI commands (logs, delete, suspend/resume, top, check), and integration test design

## Test plan
- [ ] `go test ./pkg/cli/... -v` passes
- [ ] `go vet ./pkg/cli/...` passes
- [ ] `make arena-v2` produces binary with all commands
- [ ] `./bin/arena-v2 job --help` shows all subcommands

PR #1481, opened 2026-07-20
feat: add CLI core with submission, query, and output formatting

vicoooo26:
## Summary
- `pkg/output`: table/JSON/YAML/wide output formatting
- `pkg/cli/root.go`: root command with global flags (--kubeconfig, --namespace, etc.)
- `pkg/cli/run.go`, `submit.go`: job submission commands (CLI flags and YAML)
- `pkg/cli/list.go`, `get.go`, `status.go`: job query commands
- `pkg/cli/framework_registry.go`: framework registration (MPI, PyTorch, TensorFlow)
- `pkg/cli/lifecycle.go`: resource creation helpers (ConfigMap, TensorBoard)
- `pkg/cli/gpu.go`: GPU resource extraction utilities
- `cmd/arena-v2/main.go`: binary entry point

Part of the Arena v2 CLI series (PR 5/7). Depends on #1477, #1478, #1479, #1480.

**Note:** Review only the last commit for this PR's changes.

## Previous PR
This is a resubmission of the closed PR #1469. All review comments from #1469 have been addressed and fixed in this branch.

### Review comments addressed:
- All review comments from the closed PR #1469 have been resolved, including feedback on CLI core submission flow, query commands, output formatting, framework registry, and rollback test coverage

## Test plan
- [ ] `go test ./pkg/output/... ./pkg/cli/... ./cmd/arena-v2/... -v` passes
- [ ] `go vet ./pkg/output/... ./pkg/cli/... ./cmd/arena-v2/...` passes
- [ ] `make arena-v2` produces working binary

PR #1477, opened 2026-07-20
chore: add arena v2 Makefile targets and gitignore patterns

vicoooo26:
## Summary
- Add Makefile with v2 build, test, vet, install, and e2e targets
- Add v2-specific patterns to .gitignore (test/e2e/crds/, .claude, /arena-v2)

Part of the Arena v2 CLI series (PR 1/7).

## Previous PR
This is a resubmission of the closed PR #1465. All review comments from #1465 have been addressed and fixed in this branch.

### Review comments addressed:
- Broadened `v2-integration-test` scope to cover all v2 packages (`$(V2_PACKAGES)`) instead of only `./pkg/cli/`
- Updated CLAUDE.md rule to document `arena-v2` as the consistent build target name (the `-v2` suffix is intentional during the transition period)
- Documented `ChiefWorker` as an alias for `""` (default) in `yaml-schema.md` for backwards compatibility
- Fixed `--sync-image` path mapping in `cli-flag-mapping.md` from `sync[].<git/rsync/hdfs>.image` to `sync[].image`

## Test plan
- [ ] `make arena-v2` target is defined
- [ ] `make v2-test` target is defined
- [ ] Existing code unaffected

PR #1486, opened 2026-07-21
feat: add CLI core, extended commands, completions, examples, and tests for Arena v2

vicoooo26:
## Summary

Combines PRs #1481 (CLI core), #1482 (CLI extended), and #1483 (examples/tests) into a single branch. **Depends on #1485** (v2-2-4 combined) — this PR should be reviewed/merged after #1485.

- **CLI core**: `run`, `submit`, `get`, `list`, `delete`, `version` commands with framework registry, context resolution, `--set` override support, and output formatting (JSON/table)
- **CLI extended**: `suspend`, `resume`, `status`, `logs`, `top`, `check`, `completion` commands with RBAC integration tests
- **Examples and tests**: Example YAMLs for all frameworks, e2e test suite (dry-run, submission, deletion, formatting, validation), Makefile `v2-e2e-setup`/`v2-e2e-test` targets with portable SHA-256 detection

## Review comments addressed

- **Affinity policy without rules** (#1481): Moved validation from `buildAffinity()` to `task.Validate()` — policy now requires at least one rule, rejected at validation time
- **submit.go args indirection** (#1481): Use `args[1:]` directly instead of `cmd.Flags().Args()`
- **pluralize fallback** (#1481): Added comment documenting best-effort heuristic for unknown kinds
- **extractPods synthetic names** (#1481): Already documented as fallback of last resort with extensive comments
- **go.mod version** (#1482): Reverted `go 1.26.1` → `go 1.25.1` to match develop-v2
- **AGENTS.md/CLAUDE.md** (#1482): Addressed in v2-1 PR (#1484)
- **Unqualified image names** (#1483): All e2e test images qualified with `docker.io/library/` prefix
- **Pending features test** (#1483): Added TODO comment explaining why tests are pending
- **Portable SHA256CMD** (#1477): Restored portable `sha256sum`/`shasum` detection that v2-7 had reverted

## Test plan

- [x] `go build ./cmd/arena-v2/` passes
- [x] `go vet` passes on all v2 packages + `./test/`
- [x] `go test` passes on all v2 unit test packages
- [ ] `make v2-e2e-test` passes against a real K8s cluster


pkg/task/override.go
```diff
@@ -0,0 +1,513 @@
...
+}
+
+// applyResourceOverrides handles image, workers, gpus, gpu-type, cpus, memory,
+// shm, device, envs, data, data-dir, and config-file.
+func applyResourceOverrides(t *Task, flags map[string]interface{}) error {
+	if err := setStrFlag(flags, "image", &t.Image); err != nil {
+		return err
+	}
+
+	ensureWorker := func() {
```
cheyang, line 132: **Single-node PyTorch + `--gpus` fails validation (`worker.replicas must be > 0, got 0`). [high]**

For `--workers<=1`, `buildSubmitTask` makes PyTorch master-only (`submit.go:139` `t.Worker = nil`). When `--gpus` is then passed, `ensureWorker()` here resurrects `t.Worker = &Worker{}` with `Replicas:0`, and `Validate` rejects it — so `arena submit pytorch --gpus 1 ...` (single node, one GPU) can't be submitted.

Proven end-to-end (submit → override → defaults → validate) in `TestVerify_PR1481_PytorchSingleNodeWithGPU_Validates`:
```
CONFIRMED BUG: single-node pytorch + --gpus fails validation end-to-end: worker.replicas must be > 0, got 0
```
**Suggested fix:** for single-node PyTorch, route `--gpus`/`--cpu`/`--memory` to `Master` when `Worker == nil`, or default `ensureWorker()` to `Replicas: 1`.

test/e2e/suite_test.go
```diff
@@ -0,0 +1,68 @@
+// Package e2e_test contains end-to-end tests for the arena v2 CLI.
+//
+// The lifecycle tests in this suite are CRUD smoke tests: they verify that
+// job resources can be submitted, listed, retrieved, and deleted through the
+// arena CLI. They use placeholder images and do NOT wait for pod readiness
+// or validate training outcomes.
+package e2e_test
```
cheyang, line 7: **The e2e suite has no build tag, so `go test ./...` fails without a live cluster. [high / CI]**

Every `test/e2e/*.go` file is plain `package e2e_test` (only `pkg/cli/integration_test.go` is gated with `//go:build integration`). `BeforeSuite` calls `Fail("CRD directory not found: crds …")`, so the idiomatic `go test ./...` fails at suite setup:
```
[BeforeSuite] [FAILED] CRD directory not found: crds — run `make v2-e2e-setup` first
Ran 0 of 52 Specs … FAIL github.com/kubeflow/arena/test/e2e
```
**Suggested fix:** add a build tag (e.g. `//go:build v2e2e`, or reuse `integration`) to every `test/e2e/*.go` file and drive it from the `v2-e2e-test` Makefile target, so `go test ./...` stays green without a cluster.

pkg/client/crd.go
```diff
@@ -0,0 +1,131 @@
...
+	cached := c.mpiVersion
+	c.mu.RUnlock()
+	if cached != "" {
+		return nil // cached
+	}
+
+	versions, err := c.GetCRDVersions(ctx, "mpijobs.kubeflow.org")
+	if err != nil {
+		if errors.Is(err, ErrCRDNotFound) {
+			return fmt.Errorf("MPIJob CRD not found in cluster")
```
cheyang: **`ResolveMPIVersion` severs the `ErrCRDNotFound` sentinel. [medium]**

`GetCRDVersions` returns a wrapped `ErrCRDNotFound`, but this line re-wraps it as a bare `fmt.Errorf("MPIJob CRD not found in cluster")` with no `%w`. Callers using `errors.Is(err, ErrCRDNotFound)` to tell "operator not installed" (actionable) apart from a transient/RBAC API error can no longer distinguish them. Proven in `TestVerify_PR1479_ResolveMPIVersion_PreservesSentinel`:
```
CONFIRMED BUG: ResolveMPIVersion dropped the ErrCRDNotFound sentinel (errors.Is == false); err="MPIJob CRD not found in cluster"
```
**Suggested fix:**
```go
return fmt.Errorf("MPIJob CRD not found in cluster: %w", ErrCRDNotFound)
```

cheyang (commented):
### Review — Arena v2 (#1486)

Reviewed the full cumulative v2 tree this PR carries (base `develop-v2` is the empty init commit, so #1486 brings ~28k lines across 140 files). I re-proved the round-1 findings against this head with an additive, reproducible harness — production code untouched, and the existing unit suite stays green (`go test ./pkg/... -skip TestVerify_`). Harness + evidence + `re-verify.sh`: [`verify/pr-1486`](https://github.com/cheyang/arena/tree/verify/pr-1486) (`docs/verification/pr-1486/`).

**Fixed since round 1 — thanks for addressing these:**
- `--set` array parser (former blocker): `storages[0].name=data` now round-trips and `matrix[0][1]=x` no longer panics — the `updateSliceInParent` write-back + `maxIndex` guard do the job.
- `delete -f` now honors the YAML `namespace` via `resolveNS` (flag > YAML > kubeconfig > default).

**Still open (details inline):** single-node PyTorch + `--gpus` fails validation (high); the e2e suite isn't build-gated so `go test ./...` fails (high/CI); `ResolveMPIVersion` drops the `ErrCRDNotFound` sentinel (medium).

Each inline comment cites the exact test in the harness so you can reproduce with `bash docs/verification/pr-1486/scripts/re-verify.sh`.

cheyang:
### Follow-up — re-checked at `fcd732d4`

Thanks for the update. I re-ran the verification harness against the new head after the *"fix: upgrade Go toolchain and harden RBAC, error handling, code style"* push. The two earlier fixes still hold (`--set` array parser, `delete -f` namespace), but **the three open findings from my previous review are still reproducing** — the changes to those files were cosmetic:

| Finding | Change in `fcd732d4` | Status |
|---|---|---|
| Single-node PyTorch + `--gpus` fails validation | `override.go` only `%q` formatting; `ensureWorker()` unchanged | ❌ `worker.replicas must be > 0, got 0` |
| e2e suite not build-gated | `test/e2e/*` untouched, still no `//go:build` tag | ❌ `go test ./...` fails at BeforeSuite (`CRD directory not found`) |
| `ResolveMPIVersion` drops `ErrCRDNotFound` | `crd.go` only lowercased the string (`fmt.Errorf(...)` → `errors.New("mpijob crd not found in cluster")`) — still no `%w` | ❌ `errors.Is(err, ErrCRDNotFound) == false` |

Note the `crd.go` change slightly regresses the third one: switching to `errors.New` removes any path back to wrapping the sentinel. The one-line fix is still:
```go
return fmt.Errorf("mpijob crd not found in cluster: %w", ErrCRDNotFound)
```

Details are in the three inline comments above. All of this is reproducible without a cluster via the additive harness on [`verify/pr-1486`](https://github.com/cheyang/arena/tree/verify/pr-1486):
```
bash docs/verification/pr-1486/scripts/re-verify.sh
```
Happy to re-verify again once these are addressed.

cheyang:
### Merge-blocking checklist (as of `fcd732d4`)

Summarizing what I'd consider blocking vs. non-blocking for this PR. The former blocker (`--set` array parser) is fixed; the items below remain.

#### 🔴 Should block merge

- [ ] **e2e suite is not build-gated (CI breakage).** `test/e2e/*.go` are plain `package e2e_test` with no `//go:build` tag, so `go test ./...` fails at `BeforeSuite` (`CRD directory not found: crds`). This turns CI red for the whole repo and will block unrelated PRs too. *Fix:* add `//go:build v2e2e` (or reuse `integration`) to every `test/e2e/*.go` and drive it from the `v2-e2e-test` Makefile target.
- [ ] **Single-node PyTorch + `--gpus` cannot be submitted.** `arena submit pytorch --gpus 1 …` fails with `worker.replicas must be > 0, got 0` — a core single-node training path is broken. *Fix:* for single-node PyTorch, route `--gpus`/`--cpu`/`--memory` to `Master` when `Worker == nil`, or default `ensureWorker()` to `Replicas: 1`.

#### 🟡 Recommended before merge (quality, not hard-blocking)

- [ ] **`ResolveMPIVersion` drops the `ErrCRDNotFound` sentinel** (`errors.Is == false`), so callers can't distinguish "operator not installed" from a transient/RBAC error. The `fcd732d4` switch to `errors.New` makes this harder to fix. *One-liner:* `return fmt.Errorf("mpijob crd not found in cluster: %w", ErrCRDNotFound)`.

#### ✅ Not blocking

- `delete -f` YAML namespace — **fixed**.
- `tensorboard=false` → non-nil config, and negative-priority override — latent, not reachable via the CLI today.
- MPI log selector (`replica-type` vs `job-role`) — a canary pending confirmation of the mpi-operator label contract; not asserting a fix yet.

All items are reproducible without a cluster via the additive harness on [`verify/pr-1486`](https://github.com/cheyang/arena/tree/verify/pr-1486): `bash docs/verification/pr-1486/scripts/re-verify.sh`. See the three inline comments above for per-finding detail and evidence.

cheyang:
@gemini-code-assist /review

test/e2e/suite_test.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var arenaV2Bin string
+
+func TestArenaV2(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Arena v2 E2E Suite")
+}
+
+var _ = BeforeSuite(func() {
```
cheyang, line 28: The `test/e2e` package has no `//go:build` tag, so `go test ./...` compiles and runs the Ginkgo BeforeSuite. That suite fails immediately because the `crds/` directory is missing unless `make v2-e2e-setup` has been run first. This turns the default CI test target red and will block unrelated PRs.

pkg/task/override.go
```diff
@@ -0,0 +1,513 @@
...
+		if !ok {
+			return fmt.Errorf("flag %q: expected int, got %T", "workers", v)
+		}
+		if n > 0 {
+			t.Worker.Replicas = n
+		}
+	}
+
+	// Resources
+	if gpus, ok := flags["gpus"].(int); ok && gpus > 0 {
```
cheyang, line 173: For PyTorch, `buildSubmitTask` drops the Worker block and uses Master-only mode when `--workers=1`. However, `applyResourceOverrides` still maps `--gpus` (and `--cpus`/`--memory`) to the Worker role via `ensureWorker()`, creating a Worker with `Replicas: 0`. `validateFramework` then rejects the task with `worker.replicas must be > 0, got 0`.

pkg/client/crd.go
```diff
@@ -0,0 +1,131 @@
...
+	cached := c.mpiVersion
+	c.mu.RUnlock()
+	if cached != "" {
+		return nil // cached
+	}
+
+	versions, err := c.GetCRDVersions(ctx, "mpijobs.kubeflow.org")
+	if err != nil {
+		if errors.Is(err, ErrCRDNotFound) {
+			return errors.New("mpijob crd not found in cluster")
```
cheyang: `ResolveMPIVersion` detects `ErrCRDNotFound` from `GetCRDVersions` but returns a new plain error (`errors.New("mpijob crd not found in cluster")`) instead of wrapping the sentinel. Callers that use `errors.Is(err, client.ErrCRDNotFound)` to distinguish missing operators from transient/RBAC errors will get `false`.

go.mod
```diff
@@ -1,3 +1,61 @@
 module github.com/kubeflow/arena
 
-go 1.25.1
+go 1.26.5
```
cheyang, line 3: The module declares `go 1.26.5`. If GitHub-hosted runners do not yet offer this Go release, `actions/setup-go@v5` will fail and block all CI on this PR and subsequent ones.

pkg/cli/submit_test.go
```diff
@@ -0,0 +1,481 @@
...
+	submitName = "my-job"
+	submitImage = ""
+
+	err := submitCmd.RunE(submitCmd, []string{"pytorch"})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "validation failed")
+	assert.Contains(t, err.Error(), "image is required")
+}
+
+func TestBuildSubmitTask_PyTorchWorkersNMinusOne(t *testing.T) {
```
cheyang, line 110: `TestBuildSubmitTask_PyTorchWorkersNMinusOne` covers worker counts but not the combination of `--workers=1` with `--gpus`/`--cpus`/`--memory`. A regression test would have caught the validation failure described above.

cheyang:
### Re-verified at `4b4ce3cf`

Thanks for the quick turnaround — re-ran the harness against the latest head. Two of the open findings are now fixed, one remains:

- ✅ **`ResolveMPIVersion` sentinel** — `crd.go` now wraps `fmt.Errorf("mpijob crd not found in cluster: %w", ErrCRDNotFound)`; `errors.Is(err, ErrCRDNotFound)` is true again. Confirmed by `TestVerify_PR1479_*`.
- ✅ **Single-node PyTorch + `--gpus`** — reworked `override.go` now validates end-to-end; `arena submit pytorch --gpus 1 …` passes. Confirmed by `TestVerify_PR1481_*`.

#### 🔴 Still blocking: e2e suite is not build-gated

This one is *almost* there but not wired up. `suite_test.go` dropped the crds-dir `Fail`, but:
- `BeforeSuite` still calls `Fail("arena-v2 binary not found …")`, so it still hard-fails; and
- **no `test/e2e/*.go` declares `//go:build v2e2e`** — yet `Makefile`'s `v2-e2e-test` already passes `-tags v2e2e`. The tag has nothing to match, so it's a no-op:
  ```
  $ git grep -c 'go:build v2e2e' test/e2e   # -> 0
  $ go test ./test/e2e/            # FAIL @ BeforeSuite
  $ go test -tags v2e2e ./test/e2e/ # identical FAIL — tag does nothing
  ```
  So `go test ./...` still fails. (The new `integration.yaml` only avoids it because `make v2-test` runs `go test ./test/` without `...`, which doesn't recurse into `test/e2e`.)

**One-step fix:** add `//go:build v2e2e` as the first line of every `test/e2e/*.go`. The Makefile side is already correct, so this immediately makes `go test ./...` green while `make v2-e2e-test` still runs the suite.

Everything reproducible without a cluster via [`verify/pr-1486`](https://github.com/cheyang/arena/tree/verify/pr-1486): `bash docs/verification/pr-1486/scripts/re-verify.sh`.

Non-blocking leftovers unchanged: `tensorboard=false`→non-nil config and negative-priority override (both latent / not CLI-reachable), and the MPI `replica-type` vs `job-role` selector (canary, pending mpi-operator label confirmation).

pkg/task/override.go
cheyang: ✅ Verified fixed at `4b4ce3cf`. The new `ensureResourceTarget()` routes `--gpus`/`--cpus`/`--memory` to `Master.Resources` when `Worker == nil` (single-node PyTorch), and otherwise creates `Worker{Replicas: 1}` — exactly the right fix. `arena submit pytorch --workers 1 --gpus 1 …` now passes submit → override → validate end-to-end.

pkg/client/crd.go
cheyang: ✅ Verified fixed at `4b4ce3cf`. `ResolveMPIVersion` now returns `fmt.Errorf("mpijob crd not found in cluster: %w", ErrCRDNotFound)`, so `errors.Is(err, ErrCRDNotFound)` is `true` again.

go.mod
cheyang: Walking this one back after checking: `go1.26.5` is a published **stable** release on go.dev/dl (`"stable": true`), and `actions/setup-go@v5` with `go-version-file: go.mod` installs exact patch versions from there — confirmed by pulling `go1.26.5` on a clean box via `GOTOOLCHAIN=auto`. So this shouldn't block CI; treating it as a non-issue — feel free to resolve.

pkg/cli/submit_test.go
cheyang: Still applies at `4b4ce3cf`: the production fix landed but has no unit regression test. `TestApplyOverrides_GPUs` only exercises the bare-`Task{}` branch (which auto-creates a Worker); nothing covers the single-node-PyTorch path this fix added — `Worker == nil` + `Master != nil` + `--gpus` routing to `Master`, nor `--workers=1` + `--gpus` run through `Validate`. Worth a small table case so the fix can't silently regress.

cheyang:
### Pre-merge status — verified at `d58a7a77`

Ran a full sweep (build + vet + unit suite + reproducible harness). `go build ./...`, `go vet ./...`, and the whole `go test ./pkg/...` unit suite are green. **4 of the 5 real findings are fixed — one blocker remains.**

| Finding | Status |
|---|---|
| `--set` array parser (former blocker) | ✅ fixed |
| `delete -f` honors YAML namespace | ✅ fixed |
| single-node PyTorch + `--gpus` validates | ✅ fixed |
| `ResolveMPIVersion` preserves `ErrCRDNotFound` (`%w`) | ✅ fixed |
| **e2e suite not build-gated** | ❌ **still open (blocker)** |

#### 🔴 The one remaining blocker

`test/e2e/*.go` still carry no `//go:build v2e2e` tag, so `go test ./...` fails at `BeforeSuite` (`arena-v2 binary not found` / missing `crds`). The `Makefile` already passes `-tags v2e2e` in `v2-e2e-test`, so the tag has nothing to match and is a no-op:
```
$ git grep -c 'go:build v2e2e' test/e2e   # -> 0
$ go test ./test/e2e/                          # FAIL @ BeforeSuite
```
**One-step fix:** add `//go:build v2e2e` as the first line of every `test/e2e/*.go`. That turns `go test ./...` green while `make v2-e2e-test` still runs the suite. After that, from a review standpoint there are **no remaining blockers** and this is ready for `lgtm`.

Non-blocking follow-up (already noted inline): add a unit regression test for the single-node PyTorch + `--gpus` fix so it can't silently regress.

All results are reproducible without a cluster via the additive harness on [`verify/pr-1486`](https://github.com/cheyang/arena/tree/verify/pr-1486): `bash docs/verification/pr-1486/scripts/re-verify.sh`.

vicoooo26:
> ### Pre-merge status — verified at `d58a7a77`
> Ran a full sweep (build + vet + unit suite + reproducible harness). `go build ./...`, `go vet ./...`, and the whole `go test ./pkg/...` unit suite are green. **4 of the 5 real findings are fixed — one blocker remains.**
> 
> Finding	Status
> `--set` array parser (former blocker)	✅ fixed
> `delete -f` honors YAML namespace	✅ fixed
> single-node PyTorch + `--gpus` validates	✅ fixed
> `ResolveMPIVersion` preserves `ErrCRDNotFound` (`%w`)	✅ fixed
> **e2e suite not build-gated**	❌ **still open (blocker)**
> #### 🔴 The one remaining blocker
> `test/e2e/*.go` still carry no `//go:build v2e2e` tag, so `go test ./...` fails at `BeforeSuite` (`arena-v2 binary not found` / missing `crds`). The `Makefile` already passes `-tags v2e2e` in `v2-e2e-test`, so the tag has nothing to match and is a no-op:
> 
> ```
> $ git grep -c 'go:build v2e2e' test/e2e   # -> 0
> $ go test ./test/e2e/                          # FAIL @ BeforeSuite
> ```
> 
> **One-step fix:** add `//go:build v2e2e` as the first line of every `test/e2e/*.go`. That turns `go test ./...` green while `make v2-e2e-test` still runs the suite. After that, from a review standpoint there are **no remaining blockers** and this is ready for `lgtm`.
> 
> Non-blocking follow-up (already noted inline): add a unit regression test for the single-node PyTorch + `--gpus` fix so it can't silently regress.
> 
> All results are reproducible without a cluster via the additive harness on [`verify/pr-1486`](https://github.com/cheyang/arena/tree/verify/pr-1486): `bash docs/verification/pr-1486/scripts/re-verify.sh`.

The e2e and unit tests have been organized into separate directories, and both targets have been added to the Makefile. Do we still need to run `go test` directly?

test/e2e/suite_test.go
```diff
@@ -0,0 +1,94 @@
...
+	localBin := filepath.Join("..", "..", "bin", "arena-v2")
+	if p, err := exec.LookPath(localBin); err == nil {
+		arenaV2Bin = p
+	} else if p, err := exec.LookPath("arena-v2"); err == nil {
+		arenaV2Bin = p
+	} else {
+		Fail("arena-v2 binary not found in bin/ or PATH — run `make arena-v2` first")
```
cheyang, line 35: `BeforeSuite` calls `Fail` when the `arena-v2` binary is not present, so a plain `go test ./...` (or any standard CI `go test` invocation) exits non-zero before compilation is even validated. The previous review flagged this as a build-gating blocker, and it is still present at the current SHA.

.github/workflows/integration.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Run go vet check
+        run: |
+          make v2-vet
+
+      - name: Run golangci-lint
+        run: |
+          make v2-lint
+
+      - name: Run Go unit tests
```
cheyang, line 47: The `Integration Test` workflow only runs `make v2-test`, which targets the v2 unit-test packages. The `test/e2e/` package and the `make v2-e2e-test` target are not exercised, so the PR's e2e tests are not validated in CI at all.

go.mod
cheyang, line 3: The module declares `go 1.26.5`. This is ahead of most contributor environments and may block local development or linting until Go 1.26 is widely available.

pkg/cli/lifecycle.go
```diff
@@ -0,0 +1,363 @@
...
+		Controller:         ptrBool(true),
+	}
+
+	createdResources := make([]createdResource, 0)
+
+	// SEC-7: Warn about env vars that look like secrets but are stored in
+	// plaintext in the ConfigMap.
+	for k, v := range t.Envs {
+		if isLikelySecretKey(k) && v.Value != "" && v.Secret == nil {
+			log.Warning("env var looks like a secret but is stored in plaintext in ConfigMap",
```
cheyang, line 119: `finalizeJobResources` marshals the whole `task.Task` into a ConfigMap, which includes any plaintext env vars. A warning is logged for secret-looking keys, but the value is still persisted unencrypted.

test/e2e/suite_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+var arenaV2Bin string
+
+func TestArenaV2(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Arena v2 E2E Suite")
+}
+
+var _ = BeforeSuite(func() {
```
cheyang, line 28: Twenty-three of the sixty e2e specs require a Kubernetes cluster with CRDs installed. Is the intention to add a KinD or self-hosted cluster setup to the integration workflow, or to keep those specs as manual-only tests?

pkg/cli/lifecycle.go
```diff
@@ -0,0 +1,363 @@
...
+) ([]createdResource, error) {
+	rbacResources, err := p.BuildRBAC(t, metav1.OwnerReference{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to build RBAC resources: %w", err)
+	}
+
+	var created []createdResource
+	for _, r := range rbacResources {
+		r.SetNamespace(t.Namespace)
+		r.SetOwnerReferences(nil)
```
cheyang, line 73: **Nit / low — owner-less RBAC can leak on a hard crash.** `preCreateRBAC` intentionally creates the RBAC resources with no `ownerReferences` (here), and `finalizeJobResources` backfills the owner via `PatchOwnerReferences` only after the CRD exists. The normal error paths are handled well — `submitCRD` calls `rollback(rbacResources)` if the CRD `Create` fails, and rolls back + deletes the CRD if `finalizeJobResources` fails — so this isn't a functional bug.

The only gap is a hard interruption (SIGKILL / panic / OOM) in the window between `preCreateRBAC` returning and the ownerRef patch landing: the SA/Role/RoleBinding are left with no owner, so Kubernetes GC will never reclaim them, and a later re-submit only *updates* them (via `CreateOrUpdate`) rather than noticing the orphan. Inherent to the (reasonable) "RBAC-before-CRD" ordering — not asking for a redesign, just suggesting a one-line doc caveat on this function noting the owner-less window, or an optional startup sweep for owner-less arena RBAC.

_(Verified the rest of this delta is solid: `rollback` resolves SA/Role/RoleBinding via `kindToGVR`, and the `CreateOrUpdate` conflict-retry works — apimachinery v0.36.2 `IsConflict` unwraps the `%w`-wrapped error. No other findings here.)_

test/e2e/suite_test.go
cheyang: Good question — the directory + Makefile split is clean. The catch: it doesn't change `go test ./...`, which still recurses into `test/e2e/` and hard-`Fail`s at `BeforeSuite` (no binary/CRDs). Confirmed still red at `f3a89928`.

You don't need to run it — you have the make targets. It matters for everyone *else* on the Go idiom: IDEs (run package tests on save), new contributors, generic CI/pre-commit. A repo where `go test ./...` fails out of the box surprises people.

Two low-cost fixes for a green `go test ./...` while keeping your e2e target:
1. **Build tag** — `//go:build v2e2e` on line 1 of each `test/e2e/*.go` + re-add `-tags v2e2e` to `v2-e2e-test` (you already had this). Compile-time skip under `./...`.
2. **Skip-not-Fail** — `BeforeSuite` calls `Skip(...)` instead of `Fail(...)` when infra is absent.

Not a hard blocker if maintainers standardize on `make` — just a common footgun. Happy to re-verify once you pick one.

cheyang (approved):
/lgtm
/approve

vicoooo26:
> New changes are detected. LGTM label has been removed.

The latest change only removes `main.go` from the project root.

cheyang (approved):
/lgtm
/approve

PR #1476, opened 2026-07-19
fix: prevent nil pointer dereference in servingJob and lwsJob methods

vishalmore90:
## Purpose of this PR

Fixes #1454

This PR resolves a nil pointer dereference panic occurring when viewing or listing distributed serving jobs (or any serving job where `deployment`, `lws`, or `Spec.Replicas` is `nil`).

**Proposed changes:**

- Added defensive `nil` pointer checks in `servingJob` base methods (`pkg/serving/serving.go`) when `s.deployment` is `nil`.
- Added defensive `nil` pointer checks in `lwsJob` methods (`pkg/serving/serving_distributed.go`) when `s.lws` is `nil`.
- Added `int32PtrVal` helper in `pkg/serving/util.go` to safely dereference `*int32` pointers (`Spec.Replicas`) without panicking when `Replicas` is managed by an autoscaler (e.g. HPA/KEDA).
- Updated `serving.go`, `serving_distributed.go`, and `serving_kserve.go` to use `int32PtrVal`.
- Added unit tests in `pkg/serving/serving_test.go` covering `servingJob` and `lwsJob` under nil pointer scenarios.

### Rationale

When running `arena serving get <job-name> --type distributed-serving`, `ServingJobClient.GetAndPrint()` invokes `job.GetLabels()` and other interface methods. `lwsJob` sets `deployment: nil` in its embedded `servingJob`. Without defensive nil checks, accessing `s.deployment.Labels` or dereferencing `*s.lws.Spec.Replicas` causes a CLI panic.

## Verification

Ran `go test -v ./pkg/serving/...`:

PR #1474, opened 2026-07-17
docs: update Release v0.15.5 changelog

vicoooo26:
## Summary

Add changelog entry for PR #1473 (mount init container ConfigMap into pytorch-operator) to the v0.15.5 release notes.

## Changes

- Add `feat(arena-artifacts)` entry to CHANGELOG.md referencing PR #1473

## Related Issues

Refs #1473

## Test Plan

- Verify the changelog entry renders correctly and the PR link is valid

PR #1488, opened 2026-07-27
chore(v1-release): complete v0.15.5 changelog and enable manual release trigger

vicoooo26:
## Purpose of this PR

Prepare for re-releasing v0.15.5 by backfilling missing changelog entries and adding a manual trigger to the Release workflow.

**Proposed changes:**

- Add missing v0.15.5 Bug Fixes entries for #1445, #1475, #1476 — these landed on master shortly before the changelog patch (#1474) but were not included in the v0.15.5 section.
- Add `workflow_dispatch` trigger to `.github/workflows/release.yaml` so maintainers can manually re-run a release for the current VERSION without bumping the version (e.g. re-publishing a draft after fixing its changelog).

## Change Category

- [x] Documentation update
- [x] CI/Release tooling

### Rationale

The v0.15.5 Draft Release was built from an intermediate commit. Three bug fixes (#1445, #1475, #1476) merged within minutes before the changelog patch (#1474) were omitted from the v0.15.5 section. Additionally, the Release workflow only fires on `push.paths: [VERSION]`, leaving no way to re-trigger a release for an unchanged version — which is exactly what is needed to re-publish the v0.15.5 draft with corrected artifacts and release notes.

## Test Plan

- [x] Verified `v0.15.4..release-v0.15.5` commit range matches CHANGELOG entries (all 25 user-facing PRs accounted for; only release-mechanics #1457 and changelog-patch #1474 correctly excluded).
- [x] Verified `release-v0.15.5` is a clean fast-forward descendant of `origin/master`.
- [ ] Trigger `workflow_dispatch` on a fork and confirm the Release workflow builds installer tarballs, pushes the multi-arch image, creates the `v0.15.5` tag, and creates the Draft Release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,7 +17,29 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check_tag:
+    runs-on: ubuntu-latest
+    outputs:
+      exists: ${{ steps.check.outputs.exists }}
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v7
+
+      - name: Check whether release tag already exists upstream
+        id: check
+        run: |
+          TAG="v$(cat VERSION)"
+          if git ls-remote --tags origin "refs/tags/${TAG}" | grep -q "refs/tags/${TAG}"; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+            echo "Tag ${TAG} already exists -- release is immutable, downstream jobs will skip."
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Tag ${TAG} does not exist yet."
+          fi
+
   package-arena-installer:
+    needs: check_tag
+    if: needs.check_tag.outputs.exists != 'true' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-'))
     runs-on: ubuntu-latest
 
     strategy:
@@ -56,6 +78,8 @@ jobs:
 
   build-arena-image:
     name: Build Arena container image
+    needs: check_tag
+    if: needs.check_tag.outputs.exists != 'true' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-'))
 
     runs-on: ubuntu-latest
 
@@ -199,6 +223,10 @@ jobs:
       - name: Create and push tag
         run: |
           TAG="v${VERSION}"
+          if git ls-remote --tags origin "refs/tags/${TAG}" | grep -q "refs/tags/${TAG}"; then
+            echo "Tag ${TAG} already exists upstream -- skipping tag creation."
+            exit 0
+          fi
           git tag -a ${TAG} -m "Release v${VERSION}"
           git push origin ${TAG}
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,6 +6,7 @@ on:
       - master
     paths:
       - VERSION
+  workflow_dispatch:
 
 env:
   IMAGE_REGISTRY: ghcr.io
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,10 @@
 
 ### Bug Fixes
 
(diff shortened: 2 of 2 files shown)


cheyang:
@gemini-code-assist /review

CHANGELOG.md
```diff
@@ -4,7 +4,10 @@
 
 ### Bug Fixes
 
+- fix: prevent nil pointer dereference in servingJob and lwsJob methods ([#1476](https://github.com/kubeflow/arena/pull/1476) by [@vishalmore90](https://github.com/vishalmore90))
```
cheyang, line 7: The v0.15.5 section date is still 2026-07-14, but the original release commit (#1457) landed on 2026-07-16 and these bug fixes were merged later on 2026-07-21. Since the GitHub release/tag doesn't appear to be published yet, consider updating the section date to the actual release date before shipping.

.github/workflows/release.yaml
```diff
@@ -6,6 +6,7 @@ on:
       - master
     paths:
       - VERSION
+  workflow_dispatch:
```
cheyang, line 9: The motivation checks out — only `v0.15.4` is tagged upstream, so v0.15.5 genuinely never completed and the first dispatch will work. Two things worth tightening before it goes in, though.

**`push_tag` isn't idempotent.** It runs a bare `git tag -a v${VERSION} && git push origin ${TAG}` with no existence check, and `draft_release` has `needs: push_tag`. So on a *second* dispatch — the re-release case this trigger is named for — `git tag -a` aborts with `fatal: tag already exists` (exit 128) and no release is produced, while `build-arena-image` and `release-image` have already re-pushed `ghcr.io/kubeflow/arena:${VERSION}`. The result is a mutated published image and no release, which is worse than the current failure mode. A `git ls-remote --tags origin "refs/tags/${TAG}"` short-circuit is enough to make a re-run safe.

**The new trigger isn't scoped.** `push` is limited to `master` plus `paths: VERSION`, but `workflow_dispatch` has no such restriction, and `draft_release` uses `target_commitish: ${{ github.sha }}`. A dispatch from any other branch carrying this workflow would tag and publish from that ref. `if: github.ref == 'refs/heads/master'` on the jobs would close it.

Separately, on the changelog: the date now reads 2026-07-28, which is today, but the tag still doesn't exist — so it's a prediction rather than a release date and will need another touch whenever the dispatch actually runs. Deriving it at release time, or leaving it unreleased until the tag lands, would avoid the third edit.

cheyang: Following up on the scoping half of this, with the branch list checked rather than assumed.

`master` and `release-*` are both real dispatch targets, but `develop-v2` cannot be one:

| branch | `release.yaml` | `VERSION` |
|---|---|---|
| `master` | yes | yes (`0.15.5`) |
| `release-0.12` | yes | yes (`0.12.0`) |
| `release-v0.6.0` | no (predates it) | yes |
| `develop-v2` | **no** | **no** |

The workflow file does not exist on `develop-v2`, so it never appears in the dispatch branch picker, and even if it did, `cat VERSION` would fail and there is no `arena-installer` target there. A v2 release will need its own workflow anyway, since the artifacts differ. So `master` + `release-*` covers everything that is actually dispatchable:

```yaml
jobs:
  package-arena-installer:
    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-')
    runs-on: ubuntu-latest

  build-arena-image:
    if: github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-')
    name: Build Arena container image
```

Only those two jobs need it — `release-image`, `push_tag` and `draft_release` all skip via their `needs:` chain, so there is no fifth place to forget.

One thing worth noting if you take the `release-*` half: it makes the missing tag guard more important, not less. `release-0.12` carries `VERSION=0.12.0`, so a dispatch from that branch would rebuild and overwrite `ghcr.io/kubeflow/arena:0.12.0` — an image published two years ago — and then fail at `git tag -a v0.12.0`, producing no release. Widening the allowed refs widens which published images a mistaken dispatch can clobber. Worth landing the `git ls-remote --tags` short-circuit in the same change rather than after it.

vicoooo26: A `check_tag` job has been added to the release workflow to ensure tag idempotency and prevent duplicate image builds and pushes.

cheyang (commented):
/lgtm
/approve

PR #1489, opened 2026-07-27
docs: add v2 documentation guides and finalize examples

vicoooo26:
## Summary

Add complete v2 documentation suite and finalize all example YAMLs with E2E dry-run verification. All Markdown YAML examples are now CPU-runnable with inline Python scripts and GPU resources commented out.

## Changes

- **New docs**: quickstart, installation, YAML schema, frameworks, CLI reference, best practices, monitoring, FAQ, migration, advanced
- **Examples reorganized**: directory restructure with README index, 15+ Kubeflow trainer examples converted to Arena v2 format
- **CPU-runnable**: all `train.py` references replaced with self-contained inline Python scripts, GPU resources commented out (except DeepSpeed)
- **Test fixes**: updated `examplesDir` path and replica count expectations to match reorganized examples
- **Verified**: all 20+ example YAMLs and all Markdown YAML code blocks pass `--dry-run` validation

## Test Plan

- [x] `go vet` passes on all v2 packages
- [x] `go test` passes on all v2 packages (including updated test expectations)
- [x] All example YAMLs pass `arena-v2 job run -f <file> --dry-run`
- [x] All Markdown YAML code blocks pass `--dry-run` validation
- [ ] Reviewer: spot-check a few docs pages and example YAMLs for accuracy


cheyang:
@gemini-code-assist /review

docs/best-practices.md
```diff
@@ -0,0 +1,471 @@
...
+```
+
+### Use shared memory for multi-process training
+
+Multi-process frameworks (PyTorch DDP, DeepSpeed) rely on `/dev/shm` for inter-process communication. Add a `shm` storage entry to mount a tmpfs-backed shared memory volume instead of relying on the default 64 MB.
+
+```yaml
+storages:
+  - name: shm
+    mount_path: /dev/shm     # Optional. Default: /dev/shm
```
cheyang: The documentation states that storages[].mount_path and storages[].shm are optional with defaults (/dev/shm and 2Gi), but task.Storage.Validate() in pkg/task/types.go rejects any storage with an empty mount_path, and provider.BuildVolumes only defaults the mount path when shm is non-empty but never applies a default size. A user copying the documented snippet without mount_path will get storage "shm": mountPath must not be empty.

docs/frameworks.md
```diff
@@ -0,0 +1,309 @@
...
+    memory: "2Gi"
+```
+
+Here workers handle the training computation (add GPUs when available), while the launcher runs CPU-only with modest resources — it only orchestrates the `mpirun` command and does not participate in computation.
+
+If you omit the `launcher` section, the launcher defaults to CPU-only. You can also set `framework.options.run_launcher_as_worker: true` to make the launcher inherit the worker's resources and environment variables instead.
+
+### DeepSpeed and Horovod
+
+DeepSpeed and Horovod are first-class framework names that map to the MPI provider. Setting `framework.name` to `deepspeed` or `horovod` generates an identical MPIJob CRD — the difference is purely semantic, helping you document which training engine the job uses.
```
cheyang, line 274: frameworks.md tells users that framework.name: deepspeed maps to the MPI provider and generates an MPIJob CRD. However, the only end-to-end verified DeepSpeed example in this PR (examples/v2/pretrain/deepspeed-bert.yaml) uses framework.name: pytorch and produces a PyTorchJob. This contradicts the framework guide and will confuse users about which CRD/provider DeepSpeed actually uses.

docs/cli-reference.md
```diff
@@ -0,0 +1,791 @@
...
+|------|------|---------|-------------|
+| `--dry-run` | bool | `false` | Print CRD as JSON without submitting |
+
+Also inherits [global flags](#global-flags) and `--output` from `arena job`.
+
+### Examples
+
+```shell
+# Submit a PyTorch job with 2 workers and 1 GPU each
+$ arena job submit pytorch --name my-job --image pytorch/pytorch:2.3.0-cuda12.1-cudnn8-runtime \
```
cheyang: The arena submit (legacy) section documents the command as arena submit <type>, but all four examples in that section invoke arena job submit <type>. The legacy command is registered at the root as submit, not under job, so arena job submit will fail with an unknown subcommand error.

README.md
```diff
@@ -0,0 +1,85 @@
...
+```shell
+make v2-install
+```
+
+For full installation instructions including K8s operator setup, see the [Installation guide](docs/installation.md).
+
+## Quick Start
+
+```shell
+arena check                                          # verify CRDs are installed
```
cheyang, line 66: The README quickstart shows arena check, arena job run, and arena job list, but the installation guide builds and installs arena-v2. Unless the user renames the binary (mentioned as optional in README), the quickstart commands will not work.

docs/installation.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+If a CRD is missing, the output marks it with `✗` and the command exits with an error. Ensure the Training Operator is installed correctly (see step 1) and run `arena-v2 check` again.
+
+Confirm the Arena version:
+
+```shell
+$ arena-v2 version
+Arena v2
+  Version:    0.15.4
```
cheyang: The installation guide shows arena-v2 version output as Version: 0.15.4, while the quickstart shows Version: 0.1.0. These are inconsistent placeholders and may confuse users about the actual release version.

docs/best-practices.md
cheyang: Agreed, and it goes further than `mount_path`: both annotations in that snippet are wrong. Omitting `mount_path` gives `storage "shm": mountPath must not be empty`; omitting `shm` gives `must specify exactly one of pvc, shm, tmp, hostpath, configmap, or secret`, because there is no default size defined anywhere — only `DefaultSHMMountPath`.

Worth deciding which side is wrong, because the code disagrees with itself. `pkg/provider/interface.go` does default an empty SHM mount path to `/dev/shm`, and `pkg/provider/interface_test.go` asserts that it does — but `task.Storage.Validate` (pkg/task/types.go:253) rejects the input before the provider ever sees it. So that defaulting is unreachable through `LoadFromFile`, and its unit test only passes because it constructs the `Storage` struct directly instead of loading a file.

Relaxing `Validate` to allow an empty `mount_path` when `shm` is set, plus the documented 2Gi default, would make the doc true and that branch live. Just fixing the doc is fine too, but then the provider branch and its test should go.

docs/cli-reference.md
cheyang: Confirmed: `submitCmd` is registered with `rootCmd.AddCommand(submitCmd)` at pkg/cli/submit.go:449 and never on `jobCmd`. The failure is `unknown flag: --name` rather than an unknown-subcommand error, since cobra treats `submit` as a positional argument to `job`. Four examples are affected — cli-reference.md:531, 535, 539, 543.

docs/frameworks.md
cheyang: Agreed on the contradiction. For what it is worth the guide is the correct half: `pkg/provider/mpi.go:77` does accept `deepspeed`, so `framework.name: deepspeed` really does produce an MPIJob. It is the linked example that is off — `examples/v2/pretrain/deepspeed-bert.yaml` sets `framework.name: pytorch`, so the one artifact labelled "verified end-to-end" demonstrates the opposite of what the section claims.

pkg/cli/helpers_test.go
```diff
@@ -146,10 +146,10 @@ func newFakeK8sClient(t *testing.T) *client.Client {
...
 }
 
-// examplesDir resolves the path to examples/v2 relative to pkg/cli/.
+// examplesDir resolves the path to examples/v2/quickstart relative to pkg/cli/.
 func examplesDir(t *testing.T) string {
 	t.Helper()
 	wd, err := os.Getwd()
 	require.NoError(t, err)
-	return filepath.Join(wd, "..", "..", "examples", "v2")
+	return filepath.Join(wd, "..", "..", "examples", "v2", "quickstart")
```
cheyang, line 154: Narrowing this to `quickstart/` while the example set grows from 8 files to 20 means the 16 under `reference/` and `pretrain/` end up loaded by nothing.

To be clear this isn't a bug in the PR — I dry-ran all 20 and every one parses and builds a CRD. It's about keeping them that way, since they're now user-facing docs with no coverage behind them.

The `4 -> 2` and `slotsPerWorker 4 -> 1` edits in submit_test.go are the other half of the same coupling: the assertions live in tests but the inputs live in documentation, so editing an example silently moves a test expectation. Walking `examples/v2` recursively in the existing table test in list_test.go would cover the new files and decouple the two.

cheyang (approved):
/lgtm
/approve

PR #1490, opened 2026-07-27
feat(e2e): add kind cluster setup and teardown for v2 e2e tests

vicoooo26:
## Summary

Add e2e cluster setup/teardown infrastructure for Arena v2 e2e tests. Developers can now spin up a local kind cluster with training-operator pre-installed via `make v2-e2e-setup-cluster`, run the full e2e suite, and tear it down with `make v2-e2e-teardown`.

## Changes

- Add `hack/e2e-setup-cluster.sh` script and `hack/kind-config.yaml` for creating a kind cluster with training-operator
- Add `v2-e2e-setup-cluster` and `v2-e2e-teardown` Makefile targets
- Add `.github/workflows/e2e.yaml` GitHub Actions workflow for automated e2e runs
- Add reusable `template-setup-clusters/action.yaml` composite action
- Add `busyboxImage()` helper to make e2e test images configurable via `E2E_BUSYBOX_IMAGE` env var
- Add training-operator existence check in `BeforeSuite` to fail fast with a helpful message
- Add pod creation wait step in `logs_e2e_test.go` for more reliable log tests

## Test Plan

- [ ] `make v2-e2e-setup-cluster` creates a kind cluster with training-operator
- [ ] `make v2-e2e-test` runs the full e2e suite against the cluster
- [ ] `make v2-e2e-teardown` cleans up the kind cluster
- [x] `go vet ./test/e2e/...` passes
- [x] `go build ./cmd/arena-v2/` passes

### DIFF
--- a/hack/e2e-setup-cluster.sh
+++ b/hack/e2e-setup-cluster.sh
@@ -71,11 +71,21 @@ kubectl wait deploy/training-operator \
   --for=condition=available \
   --timeout "${TIMEOUT}"
 
-# 4. Pre-load busybox image to avoid docker.io rate limits
-BUSYBOX_IMAGE="${E2E_BUSYBOX_IMAGE:-docker.io/library/busybox:1.35}"
-echo "Pre-loading ${BUSYBOX_IMAGE} into kind cluster..."
-docker pull "${BUSYBOX_IMAGE}"
-kind load docker-image "${BUSYBOX_IMAGE}" --name "${KIND_CLUSTER_NAME}"
+# 4. Pre-load images into kind to avoid docker.io rate limits
+kind_load_image() {
+  local image="$1"
+  echo "Pre-loading ${image} into kind cluster..."
+  docker pull "${image}"
+  kind load docker-image "${image}" --name "${KIND_CLUSTER_NAME}"
+}
+
+kind_load_image "${E2E_BUSYBOX_IMAGE:-docker.io/library/busybox:1.35}"
+
+# Extract and pre-load the training-operator image deployed by kustomize
+TRAINER_IMAGE=$(kubectl get deploy/training-operator \
+  -n "${NAMESPACE}" \
+  -o jsonpath='{.spec.template.spec.containers[0].image}')
+kind_load_image "${TRAINER_IMAGE}"
 
 # 5. Print cluster info
 echo ""
--- a/hack/e2e-setup-cluster.sh
+++ b/hack/e2e-setup-cluster.sh
@@ -22,13 +22,22 @@ set -o nounset
 set -o pipefail
 
 # Variables (overridable via environment)
-TRAINING_OPERATOR_VERSION="${TRAINING_OPERATOR_VERSION:-v1.9.3}"
+TRAINER_REPO="${TRAINER_REPO:-https://github.com/kubeflow/training-operator.git}"
+TRAINER_REF="${TRAINER_REF:-v1.9.3}"
+INSTALL_METHOD="${INSTALL_METHOD:-kustomize}"
 K8S_VERSION="${K8S_VERSION:-1.32.3}"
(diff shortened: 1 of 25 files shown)


test/e2e/suite_test.go
```diff
@@ -26,6 +26,11 @@ func TestArenaV2(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
+	cmd := exec.Command("kubectl", "get", "deploy/training-operator", "-n", "kubeflow")
```
cheyang: This is a reasonable guard, but it lands on top of a problem this PR is well placed to fix instead: nothing under `test/e2e/` declares a build tag, so a plain `go test ./...` still compiles this suite and dies right here.

Before this PR the failure was `arena-v2 binary not found` at suite_test.go:35. With it, it's `training-operator not found` at line 31 — and it now fails *even after* `make arena-v2`, so the obvious workaround is gone too. CI doesn't notice because `make v2-test` enumerates `V2_PACKAGES` rather than `./...`.

Adding `//go:build v2e2e` as the first line of every `test/e2e/*.go` closes it. I tried it: `go test ./...` then passes every package. One thing to pair with it — `v2-e2e-test` currently runs `go test ./test/e2e/` with no `-tags v2e2e`, so the flag needs to go back at the same time or the suite stops running entirely.

test/e2e/suite_test.go
```diff
@@ -26,6 +26,11 @@ func TestArenaV2(t *testing.T) {
...
+	if err := cmd.Run(); err != nil {
+		Fail("training-operator not found in kubeflow namespace — run `make v2-e2e-setup-cluster` first")
```
cheyang: The gate hardcodes namespace `kubeflow`, but `hack/e2e-setup-cluster.sh` exposes `NAMESPACE="${NAMESPACE:-kubeflow}"`. Run the setup with `NAMESPACE=something-else` and you get a correctly provisioned cluster that this check then refuses to run against. Reading the same variable here (defaulting to `kubeflow`) keeps the two in step.

The deployment name has the same issue on the helm path: `helm install training-operator` names the release, but the chart decides what the Deployment is called, and the helm branch skips the `kubectl wait deploy/training-operator` that the kustomize branch performs.

cheyang (commented):
The e2e workflow and the `busyboxImage()` indirection are both good calls — putting the image behind an env var instead of a hardcoded regional registry is the right fix, and swapping `_ = waitCmd.Run()` for a real `Expect(...)` in `logs_e2e_test.go` is a genuine improvement.

Two things beyond the inline comments:

- `Describe("Logs", Pending, ...)` is still marked Pending, so those strengthened assertions never actually execute. Now that this PR provisions a real training-operator, is anything still blocking un-Pending them?
- busybox isn't `kind load`-ed into the cluster, so every e2e pod pulls from docker.io on each run. That tends to show up as rate-limit flakes on public runners.

Reproductions for the inline findings: https://github.com/cheyang/arena/tree/verify/pr-1488-1490

hack/e2e-setup-cluster.sh
```diff
@@ -0,0 +1,96 @@
...
+  --image "kindest/node:v${K8S_VERSION}" \
+  --config "${SCRIPT_DIR}/kind-config.yaml"
+
+# 2. Install training-operator
+case "${INSTALL_METHOD}" in
+  kustomize)
+    echo "Installing training-operator via kustomize..."
+    kubectl apply --server-side -k "${TRAINER_DIR}/manifests/overlays/standalone"
+    ;;
+  helm)
```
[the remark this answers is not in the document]
vicoooo26: Helm chart support has been removed temporarily, and it will be added back once `training-operator` chart is available.

test/e2e/suite_test.go
vicoooo26: Helm chart support has been removed because `training-operator` chart is not available yet.

cheyang (approved):
/lgtm
/approve

PR #1492, opened 2026-07-29
test(cli): cover every examples/v2 YAML, not just the quickstart subset

cheyang:
### What

Adds `TestAllExamplesBuildCRD`, which walks `examples/v2` recursively and asserts every example
loads, validates, and builds a CRD through its framework's provider.

### Why

`examplesDir()` resolves to `examples/v2/quickstart`, so the existing tests cover 4 of the 20
shipped examples. The 16 under `reference/` and `pretrain/` are loaded by nothing.

They are all valid today — I checked every one before writing this, and the test passes on
`develop-v2` as-is. The point is to keep them that way: they are user-facing documentation, so a
schema or validation change can invalidate one and nothing fails until someone copies it into
their own YAML and hits the error themselves.

This also decouples example content from test expectations a little. Right now editing an example
can silently move an assertion — the `worker.replicas` and `slotsPerWorker` expectations in
`submit_test.go` had to change in #1489 purely because the quickstart YAML changed.

### Verification

- 20/20 examples pass on current `develop-v2`
- `make v2-test`, `go vet`, `gofmt` and `golangci-lint run` (v2.12.2, 0 issues) all clean
- Confirmed the test actually fails when it should: setting `worker.replicas: 0` in one example
  fails that example's subtest

### Known limitation

It does not catch misspelled field names. The YAML decoder ignores unknown keys, so a typo drops
the value silently and only surfaces if the resulting zero value happens to fail validation
(`worker.replcias` errors because `replicas` must be `> 0`, but `storages[].mountpath` does not).
Catching that needs strict decoding, which is a bigger change than this test and felt out of
scope here.

### DIFF
--- a/pkg/cli/examples_test.go
+++ b/pkg/cli/examples_test.go
@@ -1,0 +1,79 @@
+package cli
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/kubeflow/arena/pkg/provider"
+	"github.com/kubeflow/arena/pkg/task"
+)
+
+// allExamplesDir resolves examples/v2 relative to pkg/cli/. Unlike examplesDir,
+// which is scoped to the quickstart subset used by the CRD-shape tests, this
+// covers every shipped example so none of them can rot unnoticed.
+func allExamplesDir(t *testing.T) string {
+	t.Helper()
+	wd, err := os.Getwd()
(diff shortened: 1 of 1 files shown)


cheyang:
@gemini-code-assist /review

cheyang:
Thanks for expanding example coverage. The new test walks every `examples/v2` YAML, loads it, validates it, and confirms a CRD can be built offline. All 20 examples pass, the test runs in roughly 10 ms, and it has no external dependencies, so it doesn't add flakiness. The doc comment honestly notes the limitation around unknown YAML keys. LGTM.

Issue #1434, opened 2026-04-13
Bug: GenerateHelmTemplate ignores --set-file options, breaking --config-file


Leoyzen:
### What happened?

When using `arena serve kserve` with the `--config-file` option, the deployment fails with a Helm template rendering error:

```
Error: template: kserve/templates/inferenceservice.yaml:423:36: executing "kserve/templates/inferenceservice.yaml" at <4>: invalid value; expected string
```

**Command that failed:**
```bash
arena serve kserve \
  --name=test-model \
  --image=kserve/sklearnserver:v0.11.0 \
  --model-name=iris \
  --model-uri=s3://model-bucket/iris \
  --config-file /host/path/config.yaml:/container/path/config.yaml
```

**Root Cause:**
The bug was introduced in commit `19b5133e` ("refactor: use helm lib instead of helm binary #1207") when Arena migrated from using Helm CLI to Helm Go SDK.

In `pkg/util/helm/util.go`, the `GenerateHelmTemplate()` function accepts an `options` parameter (which contains `--set-file` options generated from `--config-file`), but **never uses these options**. The function only reads values from the values file and ignores all `--set-file` parameters.

As a result:
1. Config file paths are passed as `--set-file configFiles.<hash>.config-N.content=/path/to/file`
2. These options reach `GenerateHelmTemplate()` but are completely ignored
3. The Helm template `kserve/templates/inferenceservice.yaml` expects `$configFileInfo.content` to contain the file content
4. Since no file content was read, the value is `nil`
5. The `indent` function fails when receiving `nil` instead of a string

**Affected Versions:**
- Arena v0.15.0 - v0.15.4 (all versions after commit 19b5133e)

---

**Proposed Fix:**
Add a `processSetFileOptions()` function in `pkg/util/helm/util.go` that:
1. Parses `--set-file` options from the `options` parameter
2. Reads the file content from disk
3. Merges the content into the values map at the appropriate nested keys
4. Modifies `GenerateHelmTemplate()` to call this function when options are present

This fix maintains backward compatibility and restores the `--config-file` functionality without requiring the `--helm-binary=true` workaround.

### What did you expect to happen?

The `--config-file` option should work as before, allowing users to mount configuration files into the container. The Helm template should receive the actual file content from the `--set-file` options and render successfully.

The expected behavior is that `GenerateHelmTemplate()`:
1. Receives the `--set-file` options
2. Reads the content of the files referenced in those options
3. Merges the file content into the Helm values map
[...]

PR #1495, opened 2026-08-17
fix(helm): process --set-file options in GenerateHelmTemplate

Phoenix1504e:
## Purpose of this PR

Fixes (#1434) the `arena serve kserve --config-file` Helm template rendering error (`invalid value; expected string`).

As described in the issue, the migration to the Helm Go SDK in commit `19b5133e` broke `--config-file`. The `GenerateHelmTemplate()` function accepts an `options` parameter (which contains `--set-file` options generated from `--config-file`), but never uses it. This causes the Helm template to receive `nil` instead of the file content, leading to the rendering error.

**Proposed changes:**

- Added a `processSetFileOptions()` function in `pkg/util/helm/util.go` that parses `--set-file` options, reads the referenced file contents from disk, and merges them into the values map.
- Added a `setNestedValue()` helper function to correctly set dotted keys (e.g., `configFiles.<hash>.config-N.content`) into the nested map structure.
- Modified `GenerateHelmTemplate()` to call `processSetFileOptions()` when options are present.
- Added a unit test `TestProcessSetFileOptions` in `pkg/util/helm/util_test.go` to verify the file content is correctly nested into the values map.

This restores the `--config-file` functionality without requiring the `--helm-binary=true` workaround, and maintains backward compatibility.

## Change Category

- [ ] Documentation update

### Rationale

The root cause was that the `GenerateHelmTemplate` function ignored the `options` parameter entirely after migrating from the Helm CLI to the Helm Go SDK. By processing these options the same way the Helm CLI does (reading the file content and injecting it into the values map), the template receives the expected string content and renders successfully.

### DIFF
--- a/pkg/util/helm/util.go
+++ b/pkg/util/helm/util.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -190,30 +191,88 @@ func GenerateValueFile(values interface{}) (valueFileName string, err error) {
 
 // GenerateHelmTemplate generates helm manifests with the given valuesFile.
 func GenerateHelmTemplate(name string, namespace string, valuesFile string, chartPath string, options ...string) (templateFileName string, err error) {
-	tempName := fmt.Sprintf("%s.yaml", name)
-	templateFile, err := os.CreateTemp("", tempName)
-	if err != nil {
-		return templateFileName, err
-	}
-	defer templateFile.Close()
-	templateFileName = templateFile.Name()
+    tempName := fmt.Sprintf("%s.yaml", name)
+    templateFile, err := os.CreateTemp("", tempName)
+    if err != nil {
+        return templateFileName, err
+    }
+    defer templateFile.Close()
+    templateFileName = templateFile.Name()
 
-	values, err := chartutil.ReadValuesFile(valuesFile)
-	if err != nil {
-		return templateFileName, fmt.Errorf("failed to read values from file %s: %v", valuesFile, err)
-	}
+    values, err := chartutil.ReadValuesFile(valuesFile)
+    if err != nil {
+        return templateFileName, fmt.Errorf("failed to read values from file %s: %v", valuesFile, err)
+    }
 
-	release, err := Template(name, namespace, chartPath, values)
-	if err != nil {
-		return templateFileName, fmt.Errorf("failed to generate helm manifests %s: %v", name, err)
-	}
+    // Process --set-file options to read file contents into the values map
+    if len(options) > 0 {
+        if err := processSetFileOptions(values, options); err != nil {
+            return templateFileName, fmt.Errorf("failed to process set-file options: %v", err)
+        }
+    }
 
-	_, err = templateFile.WriteString(release.Manifest)
-	if err != nil {
-		return templateFileName, fmt.Errorf("failed to write helm manifests to file %s: %v", templateFileName, err)
-	}
+    release, err := Template(name, namespace, chartPath, values)
+    if err != nil {
+        return templateFileName, fmt.Errorf("failed to generate helm manifests %s: %v", name, err)
+    }
 
-	return templateFileName, nil
+    _, err = templateFile.WriteString(release.Manifest)
+    if err != nil {
+        return templateFileName, fmt.Errorf("failed to write helm manifests to file %s: %v", templateFileName, err)
+    }
+
+    return templateFileName, nil
+}
+
+// processSetFileOptions parses --set-file options, reads the file contents,
+// and merges them into the Helm values map.
+func processSetFileOptions(values map[string]interface{}, options []string) error {
+    for _, opt := range options {
+        // Strip the --set-file prefix if it is present
+        opt = strings.TrimPrefix(opt, "--set-file ")
+        opt = strings.TrimPrefix(opt, "--set-file=")
+
+        // Split into key and file path
+        parts := strings.SplitN(opt, "=", 2)
+        if len(parts) != 2 {
+            continue // Skip malformed options
+        }
+        key := parts[0]
+        filePath := parts[1]
+
+        // Read the file content
+        content, err := os.ReadFile(filePath)
+        if err != nil {
+            return fmt.Errorf("failed to read set-file %s: %v", filePath, err)
+        }
+
+        // Merge into values map at nested keys
+        setNestedValue(values, key, string(content))
+    }
+    return nil
+}
+
+// setNestedValue sets a value in a map at a given dotted key path (e.g. "a.b.c").
+func setNestedValue(values map[string]interface{}, key string, value interface{}) {
+    parts := strings.Split(key, ".")
+    current := values
+    for i, part := range parts {
+        if i == len(parts)-1 {
+            current[part] = value
+        } else {
+            if _, ok := current[part]; !ok {
+                current[part] = make(map[string]interface{})
+            }
+            if next, ok := current[part].(map[string]interface{}); ok {
+                current = next
+            } else {
+                // Overwrite non-map types with a map to proceed
+                newMap := make(map[string]interface{})
+                current[part] = newMap
+                current = newMap
+            }
+        }
+    }
 }
 
 // GetChartName returns the name of the chart.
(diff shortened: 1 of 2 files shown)


pkg/util/helm/util.go
```diff
@@ -190,30 +191,88 @@ func GenerateValueFile(values interface{}) (valueFileName string, err error) {
...
-	values, err := chartutil.ReadValuesFile(valuesFile)
-	if err != nil {
-		return templateFileName, fmt.Errorf("failed to read values from file %s: %v", valuesFile, err)
-	}
-
-	release, err := Template(name, namespace, chartPath, values)
-	if err != nil {
-		return templateFileName, fmt.Errorf("failed to generate helm manifests %s: %v", name, err)
-	}
+    tempName := fmt.Sprintf("%s.yaml", name)
```
cheyang: The new code is indented with spaces instead of tabs, which fails `make go-fmt` and is the reason the `build-arena` CI job is currently red. Please run `gofmt -w pkg/util/helm/util.go` and commit the result.

pkg/util/helm/util.go
```diff
@@ -190,30 +191,88 @@ func GenerateValueFile(values interface{}) (valueFileName string, err error) {
...
+    return templateFileName, nil
+}
 
-	_, err = templateFile.WriteString(release.Manifest)
-	if err != nil {
-		return templateFileName, fmt.Errorf("failed to write helm manifests to file %s: %v", templateFileName, err)
-	}
+// processSetFileOptions parses --set-file options, reads the file contents,
+// and merges them into the Helm values map.
+func processSetFileOptions(values map[string]interface{}, options []string) error {
```
cheyang: The PR description says a unit test `TestProcessSetFileOptions` was added in `pkg/util/helm/util_test.go`, but the commit only modifies `pkg/util/helm/util.go` and no test file is present. For a functional bug fix like this, we need tests covering the happy path, nested key handling, and error cases (missing file, malformed option). Please add the missing test file.

cheyang: `processSetFileOptions` strips the `--set-file` prefix but then attempts to process *every* option as a key=file pair. If a caller ever passes other Helm options (e.g., `--set foo=bar`), the function will try to read `foo=bar` as a file path and either error out or corrupt the values map. Consider filtering the input slice to only entries that start with `--set-file` before parsing, or document that this function must only receive `--set-file` options.

pkg/util/helm/util.go
```diff
@@ -190,30 +191,88 @@ func GenerateValueFile(values interface{}) (valueFileName string, err error) {
...
+
+        // Merge into values map at nested keys
+        setNestedValue(values, key, string(content))
+    }
+    return nil
+}
 
-	return templateFileName, nil
+// setNestedValue sets a value in a map at a given dotted key path (e.g. "a.b.c").
+func setNestedValue(values map[string]interface{}, key string, value interface{}) {
```
cheyang: `setNestedValue` silently overwrites an existing non-map intermediate value with an empty map when the dotted path collides with a scalar already present in the values file. This could unintentionally drop user-supplied values. Consider returning an error when an intermediate key is not a map, or at least documenting this behavior.

pkg/util/helm/util.go
```diff
@@ -190,30 +191,88 @@ func GenerateValueFile(values interface{}) (valueFileName string, err error) {
...
+func processSetFileOptions(values map[string]interface{}, options []string) error {
+    for _, opt := range options {
+        // Strip the --set-file prefix if it is present
+        opt = strings.TrimPrefix(opt, "--set-file ")
+        opt = strings.TrimPrefix(opt, "--set-file=")
+
+        // Split into key and file path
+        parts := strings.SplitN(opt, "=", 2)
+        if len(parts) != 2 {
+            continue // Skip malformed options
```
cheyang: Malformed options are currently skipped with `continue`. Returning an error would surface typos or programming mistakes immediately instead of silently ignoring them.

pkg/util/helm/util.go
```diff
@@ -190,30 +191,88 @@ func GenerateValueFile(values interface{}) (valueFileName string, err error) {
...
+
+    values, err := chartutil.ReadValuesFile(valuesFile)
+    if err != nil {
+        return templateFileName, fmt.Errorf("failed to read values from file %s: %v", valuesFile, err)
+    }
+
+    // Process --set-file options to read file contents into the values map
+    if len(options) > 0 {
+        if err := processSetFileOptions(values, options); err != nil {
+            return templateFileName, fmt.Errorf("failed to process set-file options: %v", err)
```
cheyang: nit: Use `%w` instead of `%v` when wrapping errors so callers can inspect the underlying error with `errors.Is` / `errors.As`. Same applies to the new error at line 246.

cheyang (changes requested):
Thanks for the fix. The approach of reading `--set-file` contents into the values map before templating is the right direction for restoring `--config-file`. However, this round needs two blockers addressed before we can merge: the new code is space-indented and fails `make go-fmt` (which is why `build-arena` is red), and the PR description mentions a unit test that isn't actually present in the commit. Once `gofmt` is applied and `pkg/util/helm/util_test.go` is added, the remaining items below can be addressed or discussed inline.

pkg/util/helm/util.go
cheyang: `processSetFileOptions` treats every entry in `options` as a `--set-file` key=value pair. If any caller ever passes other Helm flags through this slice, the function will try to read an unrelated path and return an error. It would be safer to filter for `--set-file` prefixes explicitly and skip unrelated options.

pkg/util/helm/util.go
cheyang: nit: the newly added code is indented with spaces, while the rest of the file uses tabs. Running `gofmt -w pkg/util/helm/util.go` will align it with the project style.

cheyang (changes requested):
The fix looks correct for the reported issue, but I'm blocking on missing unit tests: the PR body mentions `TestProcessSetFileOptions` in `pkg/util/helm/util_test.go`, yet no test file is included in the diff. Please add the tests (or equivalent coverage) and run `gofmt` on the changed file.

Phoenix1504e:
@cheyang apologies for late reply, I've updated the file with the fixes suggested by you and added unit tests too, could you please review them and suggest if any other change has to be done

pkg/util/helm/util_test.go
```diff
@@ -0,0 +1,199 @@
+// Copyright 2018 The Kubeflow Authors
```
cheyang: The license header is duplicated here (both the 2018 and 2024 copyright blocks). Please trim this to a single header so it matches the rest of the project and avoids confusion.

Phoenix1504e: Thanks for catching that, pushed the fix

cheyang (commented):
Re-review after new commit `22d8237f` (Revise copyright year in util_test.go).

The duplicated license header noted in the previous review has been removed — only a single 2024 Apache-2.0 header remains in `util_test.go`. All other previously raised items (gofmt compliance, unit-test coverage for `processSetFileOptions`/`setNestedValue`) were already resolved.

**What changed since last review:** One commit that deletes the redundant 2018 license block from the test file. No functional code changes.

**Verification:**
- `gofmt -l` reports no formatting issues on either file.
- Unit tests cover happy-path, nested merge, mixed-option filtering, `=` form, missing file, malformed option, empty key, non-map intermediate, and empty segment — comprehensive.
- The fix correctly restores `--config-file` functionality by processing `--set-file` options in `GenerateHelmTemplate`, addressing issue #1434.

No new findings. Ready to merge.

Phoenix1504e:
@cheyang I had a doubt as this is my first PR in kubeflow cla/google is the standard cla we sign on google open source right? If yes I've signed it already but the check isn't detecting it
link to the said cla agreement: https://cla.developers.google.com/about/google-individual

cheyang (commented):
Re-reviewed after the license-header cleanup. The duplicated copyright header in `pkg/util/helm/util_test.go` is resolved, `gofmt` is clean, and the unit tests for `processSetFileOptions` pass. The `--set-file` handling in `GenerateHelmTemplate` correctly parses both `--set-file key=path` and `--set-file=key=path` forms, preserves existing nested values, and returns clear errors for malformed options, missing files, and non-map intermediates. No blockers from this round.

Phoenix1504e:
@cheyang Could you please approve the pr if everything is in order?

